PR #24, opened 2020-11-17
Fix up the makefile to reference go files versus the module path

InfernoJJ:


makefile
```diff
@@ -6,7 +6,7 @@ VERSION := $(shell grep -Eo '(v[0-9]+[\.][0-9]+[\.][0-9]+(-[a-zA-Z0-9]*)?)' vers
 build:
 	go fmt ./...
 	@mkdir -p ./bin/
-	CGO_ENABLED=1 go build github.com/moov-io/iso8583
+	CGO_ENABLED=1 go build *.go
```
adamdecaf, line 9: `CGO_ENABLED=1 go build github.com/moov-io/iso8583` works on a default Go install. This sounds like a problem with a developer's laptop.

adamdecaf:
We use the current setup to build projects. I'm worried `*.go` will subtly hide sub-packages from being compiled. 

When we have binaries built that'll be `go build ./cmd/server` or something.

See: https://github.com/moov-io/ach/blob/v1.5.2/makefile#L9

InfernoJJ:
I’d like the fix for the developer laptop listed up here so the rest of us
can learn. Thanks.

PR #34, opened 2020-12-02
support track2/track3, add test samples

mfdeveloper508:


adamdecaf (approved):

pkg/lib/message.go
```diff
@@ -24,8 +24,24 @@ type Iso8583Message interface {
 }
 
 // public functions of lib
+
+// NewISO8583MessageWithJson create data elements of message with specification with json specification file
+func NewISO8583Message(specification []byte) (Iso8583Message, error) {
```
alovak, line 29: Jose, as it seems to me, new methods that you added here are related to spec creation, not to the message. I would put them into the `spec` (or util) package instead of `message`. In the end it may look like this:

```go
spec := spec.NewFromJson(file)
message := NewISO8583Message(spec)
```
or 
```go
spec := spec.NewWithAttributes(x, y)
message := NewISO8583Message(spec)
```

pkg/utils/functions.go
```diff
@@ -113,6 +115,39 @@ func IsThirdBitmap(bitmap string) bool {
 	return false
 }
 
+// Get message format
+func MessageFormat(buf []byte) string {
```
alovak, line 119: What's the goal of this function? I don't see we use it in the code only in tests.

pkg/lib/lib_test.go
```diff
@@ -770,28 +771,286 @@ func TestISO8583MessageWithValidData(t *testing.T) {
...
+	message, err := NewISO8583MessageWithAttributes(jsonData, encoding)
+	assert.Nil(t, err)
+
+	byteData, err := ioutil.ReadFile(filepath.Join("..", "..", "test", "testdata", "message_with_character_ebcdic_encoding.dat"))
+	assert.Nil(t, err)
+
+	_, err = message.Load(byteData)
+	assert.Nil(t, err)
+
+	err = message.Validate()
```
alovak, line 1028: After we validate the message, I would expect us to test that we have valid/expected values in the specific fields. Like:

```go
elements := message.GetElements()
assert.Equal(t, "xxxx", elements[4].String())
``` 

Maybe not for all .dat files, but at least for one.

pkg/utils/functions.go
mfdeveloper508: The function will use command and web server in next PR

alovak: Is it possible to move it from this branch into a new one?

pkg/lib/lib_test.go
mfdeveloper508: fixed

pkg/lib/message.go
mfdeveloper508: fixed

pkg/utils/functions.go
mfdeveloper508: removed

alovak (approved):

PR #37, opened 2020-12-08
add command lines and web server

mfdeveloper508:
add command lines, web server, tests


adamdecaf:
Nice work! It matches the some of the other CLI setups you've created. Can we add an openapi.yaml to this?

mfdeveloper508:
sure, I will add soon

api/api.yml
```diff
@@ -0,0 +1,1572 @@
...
+                $ref: '#/components/schemas/ISOMessage'
+            application/xml:
+              schema:
+                $ref: '#/components/schemas/ISOMessage'
+        '400':
+          description: bad request
+          content:
+            text/plain:
+              schema:
+                type: string
```
adamdecaf, line 506: I think we'll want to use `.Problem(err)` from moov-io/base to render a JSON object with the error. Thoughts? 

https://github.com/moov-io/base/blob/master/http/server.go#L27

adamdecaf:
The API spec looks good! Here's a preview 

https://editor.swagger.io/?url=https://raw.githubusercontent.com/moov-io/iso8583/add-cmd-server/api/api.yml

api/api.yml
mfdeveloper508: fixed

PR #53, opened 2021-03-02
Add json encoding decoding

alovak:
add JSON encoding/decoding


vtolstov:
gentle ping... json is really good stuff fo iso8583 as it easy can be placed in postgresql with hstore

alovak:
@vtolstov thanks for the feedback! It would be great if you share your use case to better understand the context. Do you want to store message data or spec or both? Do you plan to convert it back to binary from json/hstore?

vtolstov:
mostly i need to store json representation of parsed incoming iso8583 message to db and also publish to kafka (other services get needed data and store another stuff to different db).
mostly i need to store only data like "F1": "xxx", "F2":"1273817493"

alovak:
@vtolstov , thanks for the clarification. Btw, which library do you use currently?

vtolstov:
this =) master branch
i'm try to decode way4 iso8583 messages

adamdecaf:
I think storage options (such as postgres, hstore) are out of scope for this project, but I'm curious to see what you would design it to be? If you're storing the JSON blob or a more structured representation.

vtolstov:
no, i need only ability to marshal full message to json each field as map["F1"] = "value" storing in db and other stuff its my territory.
so if this pr will be merged i think that its enough for me

alovak:
This PR introduces much more, and that keeps me away from merging it. I think adding about adding JSON encoding to the message only for now. 

Let me summarise as I understand your use case:

* you will define your spec (way4)
* you will unpack binary message into go struct
* you want to encode the message into JSON to put it into DB

Is that correct?

One more question, are you going to create some go service to do this or CLI or ... how do you plan to "run the application/service"? Want to understand better how you are going to build it into your workflow.

vtolstov:
i'm decode incoming message to
```
type Message struct {    
  F2   *field.Numeric    
  F3   *field.Numeric    
  F4   *field.Numeric    
  F6   *field.Numeric    
  F7   *field.Numeric    
  F10  *field.Numeric    
  F11  *field.Numeric    
  F12  *field.Numeric    
  F13  *field.Numeric    
  F14  *field.Numeric    
  F18  *field.Numeric    
  F19  *field.Numeric    
  F20  *field.Numeric    
..............
}

after that i want to marshal it to json that contains something like
{
"mti": "0120",
"F1": "xxx",
"F2": "yyyy"
}
```

after that i'm put it into db

vtolstov:
any eta for this?

vtolstov:
gentle ping

adamdecaf:
@alovak any ideas?

alovak:
Next week I hope to have JSON encoding for messages.

alovak:
closing this in favor of #63 

@vtolstov @mfdeveloper508 thoughts?

PR #61, opened 2021-06-08
Fix encodings

alovak:
This PR fixes the issue with the wrong amount of data we have to read from the input (`data`) in order to decode the field.

Initially, I mixed two separate things: the amount of data to read and the number of characters (HEX-digits, bytes, ASCII chars) encoded in the field.

How it was before:

```go
start := f.spec.Pref.Length()
end := f.spec.Pref.Length() + dataLen
raw, err := f.spec.Enc.Decode(data[start:end], f.spec.Length)
```

where `dataLen` is the number of characters.

While it works when it's all ASCII, it doesn't work when we have BCD or a mix of encodings.

Now, it works differently:

```go
// dataLen - how many characters are encoded in the prefix
dataLen, err := f.spec.Pref.DecodeLength(f.spec.Length, data)

start := f.spec.Pref.Length()

// pass all data and tell decoder how many characters it have to decode
raw, read, err := f.spec.Enc.Decode(data[start:], dataLen)

// read - how many bytes were read from "data"
```


vtolstov:
nice, i'm try to test it with my data

vtolstov (approved):

alovak:
@vtolstov released as v0.3.1

vtolstov:
big thanks

PR #63, opened 2021-06-09
add JSON encoding for the message

alovak:
This PR adds JSON encoding (no decoding yet) of the iso8583 message.

Example:

```go
message := iso8583.NewMessage(spec)
message.MTI("0100")
message.Field(2, "4242424242424242")
message.Field(3, "123456")
message.Field(4, "100")

jsonMessage, err := json.Marshal(message)
```

`jsonMessage` contains JSON with fields ordered by index/number:

```json
{
   "0":"0100",
   "1":"700000000000000000000000000000000000000000000000",
   "2":"4242424242424242",
   "3":123456,
   "4":"100"
}
```

## Concerns

My initial format for serialized message was:

```json
{
   "fields":{
      "0":{
         "value":"0100"
      },
      "1":{
         "value":"700000000000000000000000000000000000000000000000"
      },
      "2":{
         "value":"4242424242424242"
      },
      "3":{
         "value":123456
      },
      "4":{
         "value":"100"
      }
   }
}
```

1. Getting JSON object instead of value: `{value: "0100"} vs `"0100". It lets us add more meta-data later, like format, validation, etc. But the goal of JSON serialization is to get a JSON representation of the message (kind of binary to JSON conversion).
2. We return all fields in the json root object without `fields`. Having `fields` object gives us the benefit if we plan to add other objects later (e.g. spec). 
 
Would love to get some feedback on this from @vtolstov, @mfdeveloper508. But while it's in active development (we can break API) I think we are good to go with more simple format and change it if there is a need in the future.


vtolstov:
its ok for me

alovak:
@vtolstov what do you think about removing `fields` and making it even simpler?

vtolstov:
if we dont have now and in near feature additional top level keys - when we can remove fields completely. if in feature or now we can have another top level keys - we cant remove fields =).

vtolstov:
may be encode number to string ? so fields can be map[string]string 
this makes decoding to user provided struct simpler.. ?

alovak:
I think that field type (String, Number, Binary) is responsible for serialization.

You may notice that the spec and data that you shared in way4 issue, JSON serializer will return binary data as is. I'm thinking about adding `field.Binary` that will encode value into hex (or base64).

alovak:
Also, UnmarshalJSON will check the type of the value and convert it into the proper type. Let's say if in JSON value is int but in spec it's defined as String, we will convert if from int to string on decoding.

vtolstov:
> I think that field type (String, Number, Binary) is responsible for serialization.
> 
> You may notice that the spec and data that you shared in way4 issue, JSON serializer will return binary data as is. I'm thinking about adding `field.Binary` that will encode value into hex (or base64).

Yes, will be fine to add field.Binary and i think that encoding to hex will be preferred.  As i saw logs from way4 server it encodes binary fields in logs via hex encoding

vtolstov:
> Also, UnmarshalJSON will check the type of the value and convert it into the proper type. Let's say if in JSON value is int but in spec it's defined as String, we will convert if from int to string on decoding.

ok, i'm agreed .

vtolstov:
minor comments - please add codec interface
```
type Codec interface {
  Marshal(interface{}) ([]byte, error)
  Unmarshal([]byte, interface{}) error
}
```

This will bring ability to use another json codec. like https://github.com/segmentio/encoding or jsoniterator or another json library.
Default codec can be global variable that by default will contains std lib json codec

PR #67, opened 2021-06-18
Fix zero-padded numeric value 0 unpack error

bjornouderoelink:
Unpacking a Numeric field with value 0 that is left-padded with '0' returns an error from the Unpack method instead of setting the expected value 0.

**Example**
For the following specification:
```go
field.NewNumeric(&field.Spec{
	Length:      4,
	Description: "Example",
	Enc:         encoding.ASCII,
	Pref:        prefix.ASCII.Fixed,
	Pad:         padding.Left('0'),
})
```
If the value of this Numeric field is set to 0, then in the Pack method after padding will have "0000".
When unpacking the packed output however, the Unpad method in Unpack removes all characters.
The input value to the Atoi is then "" so Atoi will return an error.
The expected output is the original input value of 0 instead of an error.

This PR fixes this bug by only parsing the raw if the length is not 0, otherwise the value is set to 0.
Let me know if I should also create an issue for this.

### DIFF
--- a/field/numeric.go
+++ b/field/numeric.go
@@ -85,7 +85,7 @@ func (f *Numeric) Unpack(data []byte) (int, error) {
 	if len(raw) == 0 {
 		// for a length 0 raw, string(raw) would become "" which makes Atoi return an error
 		// however for example "0000" (value 0 left-padded with '0') should have 0 as output, not an error
-		// so if the lenght of raw is 0, set f.Value to 0 instead of parsing the raw
+		// so if the length of raw is 0, set f.Value to 0 instead of parsing the raw
 		f.Value = 0
 	} else {
 		// otherwise parse the raw to an int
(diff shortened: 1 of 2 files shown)


vtolstov:
LGTM

vtolstov (approved):

adamdecaf (approved):

alovak (approved):
Looks good to me! Thanks for the contribution, @bjornouderoelink !

alovak:
@bjornouderoelink do you want me to fix the typo and merge or you want to do it yourself?

bjornouderoelink:
Ah, a typo. 
It should be fixed now, but I wouldn't have minded if you fixed it either.

alovak:
Fix released as [v0.3.3](https://github.com/moov-io/iso8583/releases/tag/v0.3.3)

PR #68, opened 2021-06-25
Handle ISO8583 Composite Field Types

krishishah:
This pull request adds support to this library to set, pack and unpack ISO8583 Subfields. We define 'subfields' here as a set of structured fields within a parent field which can be encoded and decoded in a set order (ordered by ID). Note that this enhancement does not cater for TLVs or subelements (not to be confused with subfields). See the 'Next Steps' section below for more detail on that.

Unfortunately, this pull request is more involved than I would've ideally liked and also includes breaking changes for users of the library. 

## New Behaviour

As a consequence of this PR, users may now set, get, pack and unpack ISO8583 `Subfields` as shown below.

Firstly, users may defined `Subfields` within a `MessageSpec` as such:

```
    spec := &MessageSpec{
        Fields: map[int]field.Field{
            ...
            3: field.NewSubfields(&field.Spec{
                Length:      6,
                Description: "Processing Code",
                Pref:        prefix.ASCII.Fixed,
                Fields: map[int]field.Field{
                    1: field.NewString(&field.Spec{
                        Length:      2,
                        Description: "Transaction Type",
                        Enc:         encoding.ASCII,
                        Pref:        prefix.ASCII.Fixed,
                    }),
                    2: field.NewString(&field.Spec{
                        Length:      2,
                        Description: "From Account",
                        Enc:         encoding.ASCII,
                        Pref:        prefix.ASCII.Fixed,
                    }),
                    3: field.NewString(&field.Spec{
                        Length:      2,
                        Description: "To Account",
                        Enc:         encoding.ASCII,
                        Pref:        prefix.ASCII.Fixed,
                    }),
                },
            }),
            ...
        },
    }
```

Using the untyped API, subfields may be set and packed in a consistent manner with all other field types:

```
    err := message.Field(3, "123456")
    if err != nil {
        fmt.Errorf("...")
    }
    bytes, err := message.Pack()
```

Furthermore, subfields may also be retrieved using the untyped API in a consistent manner with all other field types:

```
    // s == "123456"
    s, err = message.GetString(3)
```

Using the typed API, `Subfields` data may be defined as such:

```
    type TestISOF3Data struct {
        F1 *field.String
        F2 *field.String
        F3 *field.String
    }

    type ISO87Data struct {
        ...
        F3 *TestISOF3Data
        ...
    }
```

And packed:    
```
    message := NewMessage(spec)		
    message.MTI("0100")		
    err := message.SetData(&ISO87Data{	
        ...
        F3: &TestISOF3Data{
	      F1: field.NewStringValue("12"),
	      F2: field.NewStringValue("34"),	
	      F3: field.NewStringValue("56"),	
        },	
    })
    rawMsg, err := message.Pack()
```

They may be unpacked and retrieved in a typed manner as such:

```
    message.SetData(&ISO87Data{})
    
    err := message.Unpack([]byte(...))
    if err != nil {
        ...
    }
    data := message.Data().(*ISO87Data)

    f1 := data.F3.F1.Value // "12"
    f2 := data.F3.F2.Value // "34"
    f3 := data.F3.F3.Value // "56"
```

JSON representation of subfields is as follows:

```
{
   "0":"0100",
   "1":"700000000000000000000000000000000000000000000000",
   "2":"4242424242424242",
   // Before this would have been a flat field as such: "3": "123456".
   // This structure mirrors that of the message spec but does vary away from 
   // just having flat fields. We should discuss if this is really the behaviour we want.
   "3": {
       "1": "12",
       "2": "34",
       "3": "56",
     }
   "4":"100"
}
```

## Changes Made 

We have amended the `field.Field` interface as such:

Old:
```
type Field interface {    
    Spec() *Spec    
    SetSpec(spec *Spec)    
    
    Pack() ([]byte, error)    
    Unpack(data []byte) (int, error)    
    
    SetBytes(b []byte)    
    Bytes() []byte    
    
    String() string
}
```

New:
```
type Field interface {    
    Spec() *Spec    
    SetSpec(spec *Spec)    

    Pack() ([]byte, error)    
    Unpack(data []byte) (int, error)
    
    SetBytes(b []byte) error    
    Bytes() ([]byte, error)    
    
    SetData(d interface{}) error    

    String() (string, error)
}
```

`SetData(...)` behaves in a similar fashion to `Message.SetData`. It leverages reflection to traverse through fields provided in the `data` arg and sets their `Spec` with the correct value. For scalar types such as `String` and `Numeric`, `SetData` simply involves storing a reference to the `data` field provided along with its `Value` (if it has been set). When `Unpack` is called (using an empty `data` field), these types then populate the deserialised value of said field into the `data` struct provided. When `Pack` is called using a populated `data` field (provided using `SetData`), its provided value is serialised into bytes. 

The eagle eyed of the audience will also note that `error` has been added to the return types of `SetBytes(...)`, `Bytes()` and `String()`. This is because our implementation of `Subfields` leverages `Unpack` and `Pack` respectively in order to set and get bytes to and from the underlying fields. More information on this can be found in the respective method documentation. Unfortunately, this results in a breaking change for users on previous versions of the library - an alternative would be to panic or drop the error, both of would lead to unexpected client observable behaviour.

To allow for fields to be nested within other fields, we also enhance the `field.Spec` struct to hold additional information. Namely:

```
type Spec struct {                                                                                                                                                          
    Length      int                                                                                                                                                             
    Description string                                                                                                                                                         
    Enc         encoding.Encoder                                                                                                                                               
    Pref        prefix.Prefixer                                                                                                                                                
    Pad         padding.Padder   
    // This may be changed to `map[string]Field` in the future to accommodate for TLVs.                                                                                                                                           
+   Fields      map[int]Field                                     
}  
```

The `Fields` field is leveraged to set `Spec`s within nested subfields or generate empty defaults if the untyped API is used. Due to nuances in the implementation of `Subelements` and `TLVs`, it's possible that we may have to change the type of `Fields` from `map[int]Field` to `map[string]Field`. I held back from making such a change in this PR to prevent premature optimisation, however it does mean more breaking changes could be made soon.

Finally, we have updated all of the constructors in the `field` package to return a pointer to their concrete types rather than the `field.Field` interface. This change is inspired by the '[Accept Interfaces Return Struct](https://mycodesmells.com/post/accept-interfaces-return-struct-in-go)' Golang principle. Interfaces are implemented implicitly in Go and this means that a type can implement multiple interfaces at once, without me doing anything except defining methods the type should use.

## Next Steps

Since we have laid down most of the boilerplate required to set, get, pack and unpack complex fields, users may now build support for `Subelements` and `TLVs`. After briefly skimming through the documentation of several card schemes, the format of `subelements` seem to vary considerably across different Data Elements and schemes. Without a standard definition for such a field type, it doesn't seem sensible to incorporate such specific field types into this open source library. However, as a result of this PR, users are now able to define bespoke `subelement` formats themselves, define them in message specs use them for packing + unpacking. In addition, TLVs and other bespoke subelement structures do not require fields to be set in any specific order (whereas subfields are always ordered by ID). Therefore, the logic required to pack and unpack these fields differs from standard subfields and as such, they must be represented by different concrete types.

Support for `TLVs` requires some additional work. TLVs hold alphanumeric/binary IDs (called tags) instead of the standard integer IDs leveraged by data elements, subfields and subelements. As described above, this may require that we make further breaking changes to the `Spec` struct once again. Following on from that, these PRs should be much more straightforward to implement given nothing unexpected arises along the way 🤞.


alovak:
@krishishah thanks a lot for your contribution! I need some time to go through the PR, but will try to get back to you shortly!

field/subfields_test.go
```diff
@@ -0,0 +1,336 @@
...
+		desc string
+		spec *Spec
+	}{
+		{
+			desc: "panics on non-None / non-nil Enc value being defined in spec",
+			spec: &Spec{
+				Length: 6,
+				Pref:   prefix.ASCII.Fixed,
+				Enc:    encoding.ASCII,
+				Pad:    padding.Left('0'),
```
alovak, line 308:
```suggestion
				Pad:    padding.None,
```

field/subfields_test.go
```diff
@@ -0,0 +1,336 @@
...
+				Fields: map[int]Field{},
+			},
+		},
+		{
+			desc: "panics on non-None / non-nil Pad value being defined in spec",
+			spec: &Spec{
+				Length: 6,
+				Pref:   prefix.ASCII.Fixed,
+				Enc:    encoding.ASCII,
```
alovak, line 317:
```suggestion
				Enc:    encoding.None,
```

alovak (commented):
@krishishah this PR looks amazing! This change is huge 💪! Thank you and your team for the contribution!

I have some small cosmetic comments... and we are good to go!

krishishah:
Thanks for the review @alovak !

I've addressed your comments and done a little refactoring in the mean time. We discovered some extra usecases we needed to accommodate for and found a way to do it generically and as such we've renamed the `Subfields` type to `Composite`

As of the most recent commit, `Composite` can pack and unpack bytes which have `IDs` embedded into the data itself. Several implementations of both ISO8583 subfields and subelements do this. These can also be unpacked out of order (another nuance that schemes do) and non-present subfields can be handled gracefully now as well.

I've introduced an additional `IDLength` field to the spec to accommodate for this usecase and overall the implementation is more complete than it was initially. Much of the documentation for this can be found in `field/composite.go`. We'd actually like to get this merged ASAP so a prompt review would be much appreciated 🙏 

Let me know what you think!

field/composite.go
```diff
@@ -0,0 +1,325 @@
...
+func (f *Composite) SetSpec(spec *Spec) {
+	if err := validateCompositeSpec(spec); err != nil {
+		panic(err)
+	}
+	f.spec = spec
+	f.idToFieldMap = spec.CreateMessageFields()
+	f.orderedSpecFieldIDs = orderedKeys(f.idToFieldMap)
+}
+
+// SetData traverses through fields provided in the data parameter matches them
```
alovak, line 90: I see that you moved code that sets data into `pack` method. It's more compact now, but in my opinion it has following drawbacks:
* if there is an error for setting data (e.g. wrong type) you will know about it during `pack`, not `SetData`
* when you call `SetData` you may want to access message fields in the struct, but values are not there until you call `pack`

Is there a reason to move the logic described in the comment into `pack` method?

wadearnold (approved):

alovak:
@krishishah I really like how you simplified code by adding `SetData` to the fields and using this construction inside:

```go
	if f.data != nil {
		*(f.data) = *f
	}
```

field/composite.go
krishishah: Yes, I moved the code into both `Pack` and `Unpack` from `SetData`. I've made the same changes in `message.go`.

You're right about the type checking drawback. Unfortunately, the issue we had previously with `SetData` was that it used reflection to perform checks and created new instances of all nil fields but we needed more granularity on when to initialise these new fields.

When we are packing fields, we actually never wish to initialise nil fields because it is up to the user to decide which fields get converted to bytes.

When unpacking fields, we wish to initialise the relevant fields whose data has been deserialised from bytes (leaving non-present fields as nil).

We've added a test to assert against this in `composite_test.go`: `"Unpack correctly deserialises partial subfields to the data struct"` line 449 where we test against a struct which should be partially populated.

This test previously failed because `F2` was populated with its default value (empty string) instead of being nil and what led me to refactoring its implementation.

Because `SetData` is shared between both the packing and unpacking usecases, we aren't able to make fine grained decisions on what type checks to make and which fields to initialise and as such we are forced to move that logic into `Pack()` and `Unpack()`

> when you call SetData you may want to access message fields in the struct, but values are not there until you call pack

I haven't quite understood this dawback though. Users should be able to access the fields of their data struct directly because they created it and may hold a reference to its pointer if they desire.

In addition, calling `String()` or `Bytes()` calls `pack()` under the hood so they should be able to access these fields via the untyped API. Is there an entry point that I'm missing out on?

alovak:
@krishishah released under `v0.4.0`. We will update docs and release notes a little bit later. Thanks for your contribution!

PR #75, opened 2021-07-06
Add network header

### LINKED ISSUE #70: Utility Pack & Unpack from network connection
A network connection to a processing network (visa) has a two-byte message length plus the unpacked or packed ISO8583 message. 

To send a message you must iso8583.pack() the iso message and check that you received the bytes without an error. Then you use [binary.size](https://golang.org/pkg/encoding/binary/#Size) to get the length of the message. You write a bugger of length size and add the size and the packed data before it is submitted to the network. 

To receive a message on a net connection you read the first two bytes, throw them away, and then unpack the bytes into an ISO message. 

I would like to see the ability to send/parse a message over a go socket connection. It seems like this has to be a common pattern for anyone using this library on a network. Based on other comments it seems like each network has a different header length. The length of the header probably needs to be configurable.

alovak:
All messages between client/server have a message length header. In some cases, it can be a 4 bytes ASCII or 2 bytes BCD encoded length. With this PR we add the interface and two implementations of the `network.Header`:

```go

// Header is the interface for network header. 
type Header interface {
	// WriteTo encoded length into Writer
	WriteTo(w io.Writer) (int, error)

	// ReadFrom reads header (N bytes) from the Reader
	ReadFrom(r io.Reader) (int, error)

	// SetLength sets the length of the message
	SetLength(length int)

	// Length returns the length of the message
	Length() int
}
```

Here is how you can use it to read message from net.Conn:

```go
func handleRequest(conn net.Conn) {
	header := network.NewBCD2BytesHeader()
	_, err := header.ReadFrom(conn)
	if err != nil {
		// handle error
	}

	// Make a buffer to hold message
	buf := make([]byte, header.Length())
	// Read the incoming connection into the buffer.
	read, err := conn.Read(buf)
	if err != nil {
		// handle error
	}
	if reqLen != header.Length() {
		// handle error
	}

	message := iso8583.NewMessage(iso8583.Spec87)
	message.Unpack(buf)

	// work with the message
}
```

This is how you can write header into the net.Conn:

```go
header := network.NewBCD2BytesHeader()
packed, err := message.Pack()
if err != nil {
	// handle error
}
header.SetLength(len(packed))
_, err = header.WriteTo(conn)
if err != nil {
	// handle error
}
n, err := conn.Write(packed)
if err != nil {
	// handle error
}

```

Closes #70

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@ ISO8583 implements an ISO 8583 message reader and writer in Go. ISO 8583 is an i
 	- [Build message](#build-and-pack-the-message)
 	- [Parse message](#parse-the-message-and-access-the-data)
 	- [JSON encoding](#json-encoding)
+	- [Network header](#network-header)
+- [ISO8583 CLI](#cli)
 - [Learn about ISO 8583](#learn-about-iso-8583)
 - [Getting help](#getting-help)
 - [Contributing](#contributing)
@@ -243,6 +245,63 @@ it will produce following JSON:
    "2":"4242424242424242",
    "3":123456,
    "4":"100"
+}
+```
+
+### Network Header
+
+All messages between the client/server (ISO host and endpoint) have a message
+length header. It can be a 4 bytes ASCII or 2 bytes BCD encoded length. We
+provide a `network.Header` interface to simplify the reading and writing of the
+network header.
+
+Following network headers are supported:
+
+* ASCII4Bytes - message length encoded in 4 bytes ASCII, e.g., 0115 for 115
+  bytes of the message
+* BCD2Bytes - message length encoded in 2 bytes BCD, e.g, {0x01, x15} for 115
+  bytes of the message
+
+You can read network header from the network connection like this:
+
+```go
+header := network.NewBCD2BytesHeader()
+_, err := header.ReadFrom(conn)
+if err != nil {
+	// handle error
+}
+
+// Make a buffer to hold message
+buf := make([]byte, header.Length())
+// Read the incoming message into the buffer.
+read, err := conn.Read(buf)
+if err != nil {
+	// handle error
+}
+if reqLen != header.Length() {
+	// handle error
+}
+
+message := iso8583.NewMessage(iso8583.Spec87)
+message.Unpack(buf)
+```
+
+Here is an example of how to write network header into network connection:
+
+```go
+header := network.NewBCD2BytesHeader()
+packed, err := message.Pack()
+if err != nil {
+	// handle error
+}
+header.SetLength(len(packed))
+_, err = header.WriteTo(conn)
+if err != nil {
+	// handle error
+}
+n, err := conn.Write(packed)
+if err != nil {
+	// handle error
 }
 ```
 
--- a/network/ascii_4bytes_header.go
+++ b/network/ascii_4bytes_header.go
@@ -1,0 +1,50 @@
+package network
+
+import (
+	"fmt"
+	"io"
+	"strconv"
+)
+
+var _ Header = (*ASCII4BytesHeader)(nil)
+
+// ASCII4BytesHeader is 4 bytes ASCII encoded length
+type ASCII4BytesHeader struct {
+	Len int
+}
+
+func NewASCII4BytesHeader() *ASCII4BytesHeader {
+	return &ASCII4BytesHeader{}
+}
+
+func (h *ASCII4BytesHeader) SetLength(length int) {
+	h.Len = length
+}
+
+func (h *ASCII4BytesHeader) Length() int {
+	return h.Len
+}
+
+func (h *ASCII4BytesHeader) WriteTo(w io.Writer) (int, error) {
+	return fmt.Fprintf(w, "%04d", h.Len)
+}
+
+func (h *ASCII4BytesHeader) ReadFrom(r io.Reader) (int, error) {
+	buf := make([]byte, 4)
+	read, err := io.ReadFull(r, buf)
+	if err != nil {
+		return 0, fmt.Errorf("reading header: %v", err)
+	}
+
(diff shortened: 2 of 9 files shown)


wadearnold:
Why would you need to break the API. The API today assumes a header of Length 0 today in both Unpack/Pack. If the header is not set then it would still have the same behaviour as it currently provides. 


The current implementation: 

```go
message := iso8583.NewMessage(iso8583.Spec87)
err := message.Unpack([]byte("0210323800010AC000080030000000000006630706101623000001101623070604008910162341603896JUNI0SIM00000000000999900504.10"))
```

If I want to deal with messages that have a header then I just add the concrete Header implementation to the instantiation after I set the spec. 

```go
message := iso8583.NewMessage(iso8583.Spec87)
message.setHeader(iso8583.NewVisaDEXHeader())
err := message.Unpack([]byte("01150210323800010AC000080030000000000006630706101623000001101623070604008910162341603896JUNI0SIM00000000000999900504.10"))
```

The default header could be of length zero and we could add a couple of common header types to the library or allow for custom via implementing the Header interface? 

We would want to expose the value of the header after it was packed or unpacked 

```go
message.Header
```

I think the interface will need to set the value of the header regardless of how it is encoded.

alovak:
Our current API supports the cases when you have both header and message in `Unpack(src []byte)`. But it works only when we got the message from somewhere else. When you read from the network, you don't know how much data you have to read. You know that you have to read N bytes (or 2 in your case) for the header, then the header will tell you how many bytes to read for the message. This can be achieved if we:
* add ISOHeader as a separate thing, read message separately, and then pass it into `message.Unpack`
* pass Reader into `message.Unpack` instead of slice and let Message read as much data as it should

wadearnold:
My point is that you are always going to know your header. You either don't have one because you are just parsing iso8583 messages for fun or you are connected to a network. If you are connected to a network you know the number of bytes in the header that needs to be read or written. If you pack a message to be sent over the network you still need to write a header that has the message length in the first N bytes. If you read a message you need to unpack the first N bytes. 

If you are going to break the unPack function why would you not just rename Unpack() -> read() and Pack()->Write() and implement it as an io.ReaderWriter interface? This is what we did in the ACH project. It has the advantage that you can send the HTTP or Socket connection straight into the library without making a copy.

alovak:
@wadearnold it was good that you mentioned Read/Write in ACH. I just added `Read` and `Write` methods for the message that accepts `io.Reader` and `io.Writer`. They will read/write headers and internally they call `Unpack()` method.

header/dex.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"io"
+	"strconv"
+
+	"github.com/moov-io/iso8583/encoding"
+)
+
+var _ Header = (*VisaDexHeader)(nil)
+
+// Visa DEX header is 2 bytes of BCD encoded length
```
wadearnold, line 13: Could we just make this a 2ByteHeader. I know Visa DEX is only two bytes but Discover and others are the same and that makes it seem strange to use the Visa name for other networks. Visa also probably crawls the internet looking for their name.

alovak: Done!

I think the next step can be to add CustomHeader:
```go
header := NewCustomHeader(CustomHeaderOptions{
  bytes: 2,
  enc: encoding.BCD,
})
```

then all other headers can be implemented via Custom.

message.go
```diff
@@ -140,6 +147,53 @@ func (m *Message) Pack() ([]byte, error) {
 	return packed, nil
 }
 
+func (m *Message) Read(reader io.Reader) error {
```
wadearnold, line 150: Why is this code not just a part of Unpack?

message.go
```diff
@@ -140,6 +147,53 @@ func (m *Message) Pack() ([]byte, error) {
...
+	} else {
+		src, err = io.ReadAll(reader)
+		if err != nil {
+			return fmt.Errorf("failed to read src from the reader: %v", err)
+		}
+	}
+	return m.Unpack(src)
+}
+
+func (m *Message) Write(writer io.Writer) error {
```
wadearnold, line 173: Why is Write and Pack not the same function?

alovak:
@wadearnold I'll keep only Write/Read as public methods. Will make changes soon.

wadearnold:
I would assume almost everyone using Unpack until now had some code to convert the string to bytes. 

```go
err := message.Unpack([]byte("0210323800010AC000080030000000000006630706101623000001101623070604008910162341603896JUNI0SIM00000000000999900504.10"))
```

If we deprecated  `Unpack()` and just made it a `Read()` you would be able to do the same thing. 

```go 
err = imessage.Read(strings.NewReader("0210323800010AC000080030000000000006630706101623000001101623070604008910162341603896JUNI0SIM00000000000999900504.10"))
```

The entire folder of a project could be updated. 

```bash
cd /path/to/your/folder
sed -i 's/.Unpack([]byte/.Read(strings.NewReader/g' *
```

I am not sure of a simple way to update all of the `Pack()` to `Write()` in everyone's codebase.

krishishah:
The current state of this looks really good. If we are deprecating `Pack()` / `Unpack()` in favour of `Write()` and `Read()` for the message interface, is the plan to follow suit with the `field.Field` interface? It'll require passing `io.Reader` down the call chain.

alovak:
@krishishah yes. I'm looking into changing Field as well

alovak:
Working on this in a separate branch

alovak:
@wadearnold I updated the PR.

I moved headers into the `network` package to signify that it's a `network` header. If we put them into `utils` then we need to add the `Network` prefix to the constructor and types to distinct it from the message header. I hope that works for you.

wadearnold (approved):

PR #76, opened 2021-07-08
Replce Pack/Unpack with ReadFrom/WriteTo

alovak:
# Changes

* we implement `io.ReaderFrom` interface by adding `ReadFrom(io.Reader)` method to the `Message`. `ReadFrom` reads data fields sequentially from the `io.Reader`
* we implement `io.WriterTo` interface by adding `WriteTo(io.Writer)` method to the `Message`. `WriteTo` writes data fields sequentially to the `io.Writer`
* we implement mentioned interfaces for all fields as well

Example:

to read from `Reader`:
```go
// reader can be a file, socket, etc.
reader := bytes.NewReader([]byte{48, 49, 48, ... })
n, err := message.ReadFrom(reader)
```

to write into `Writer`:
```go
// writer can be a file, socket, etc.
writer := bytes.NewBuffer([]byte{})
n, err := message.WriteTo(writer)
```

## Why this change is needed?

When you read messages from the network you don't know how much data to read. In the previous implementation of the package, we worked with provided message. How this message was received was out of the scope of this library. In the real-world, such messages can be read from the network or file.

The proper way is to read fields of the message one by one from the `Reader` (input). We know the length of the fields and we know when to stop reading data for the message.

It also will help us to with further work on ISO header.


wadearnold:
I have asked @adamdecaf to chime in. IMHO these should be `read` and write` so that they are composable and you don't create indirection.

field/composite.go
```diff
@@ -117,55 +119,75 @@ func (f *Composite) SetData(data interface{}) error {
 
 // Pack deserialises data held by the receiver (via SetData)
 // into bytes and returns an error on failure.
-func (f *Composite) Pack() ([]byte, error) {
-	packed, err := f.pack()
+func (f *Composite) WriteTo(w io.Writer) (n int, err error) {
```
adamdecaf, line 122: Is `Pack()` not a useful method to offer? It looks like the initial design was allocating byte slices and now you're trying to offer a streaming/buffered interface? 

Did we make this change because of potential performance problems with finite byte slices (e.g. higher memory allocations, delays populating the entire array over multiple packets)

alovak:
The first question I asked myself is: why I’m changing this at all?

the answer is: to make it easier (and properly) work with network.

I would love to hear feedback and ideas from @adamdecaf

field/composite.go
alovak: While the reason to change Unpack is properly read from network (when you may not know how much data to read)! Pack was changed for consistency. I could change only Pack in the Message and keep the rest unchanged.

field/composite.go
```diff
@@ -117,55 +119,75 @@ func (f *Composite) SetData(data interface{}) error {
...
-// Unpack takes in a byte array and serializes them into the receiver's
-// subfields. An offset (unit depends on encoding and prefix values) is
-// returned on success. A non-nil error is returned on failure.
-func (f *Composite) Unpack(data []byte) (int, error) {
-	dataLen, err := f.spec.Pref.DecodeLength(f.spec.Length, data)
+// ReadFrom takes in an io.Reader and reads data from it and then serializes
+// them into the receiver's subfields. An offset (unit depends on encoding and
+// prefix values) is returned on success. A non-nil error is returned on
+// failure.
+func (f *Composite) ReadFrom(r io.Reader) (int, error) {
```
adamdecaf, line 155: It seems like we're changing this behavior from:

`Unpack` was requiring the entire message be copied from the network even though we may only need to consume part of it initially? e.g. read message headers
```
AAAABBBBBBCCCCCC (needs to be consumed and written into []byte)
      ^ (the offset we only needed to read from)
```

`ReadFrom` looks to be trying to offer the ability to _only_ consume `AAAABB` if that's all of the message needed for now, but the [Go docs on that interface](https://golang.org/pkg/io/#ReaderFrom) (`ReaderFrom`) mention that the entire message will be consumed still. 

`Composit.ReadFrom` mentions it returns an _offset_ but Go's docs return `n` as the **total number of bytes read**. That behavior breaks the ABI but would still compile and introduce bugs. Also note Go requires it be measured in bytes but this PR's units can vary at runtime. 

Of course this could just be renaming methods so that they conform to Go's stdlib interfaces which allows for interop with lots of packages.

encoding/bcd.go
```diff
@@ -45,3 +48,31 @@ func (e *bcdEncoder) Decode(src []byte, length int) ([]byte, int, error) {
 	// e.g. 0643 => 643
 	return dst[decodedLen-length:], read, nil
 }
+
+func (e bcdEncoder) DecodeFrom(r io.Reader, length int) (data []byte, read int, err error) {
```
alovak, line 52: I don’t like this method. I’m thinking about keeping Decode only and find other way to read data for decoding.

field/composite.go
```diff
@@ -117,55 +119,75 @@ func (f *Composite) SetData(data interface{}) error {
...
+	_, err := f.unpack(bytes.NewReader(data), len(data))
 	return err
 }
 
 // Bytes iterates over the receiver's subfields and packs them. The result
 // does not incorporate the encoded aggregate length of the subfields in the
 // prefix.
 func (f *Composite) Bytes() ([]byte, error) {
-	return f.pack()
+	buf := bytes.NewBuffer([]byte{})
```
adamdecaf, line 184:
```suggestion
	var buf bytes.Buffer
```

adamdecaf: If you do need to allocate memory ahead of time there's `Buffer.Grow(int)` as well. 

https://golang.org/pkg/bytes/#Buffer.Grow

field/composite.go
adamdecaf: Pack -> WriteTo makes sense. Network writes are going to use `io.Writer`. 

We could keep Pack() and use `bytes.Buffer` to maintain backwords compatibility.

encoding/bcd.go
adamdecaf: Similar to this I was wondering why an `Encoder` interface contained "Decode" methods.

wadearnold:
I know we went down this path because of Header. Not sure how we got to here. At this point, I would rather see us keep with `Pack` and `Unpack` and revisit this when we are actually closer to the domain of a socket server and network.

field/composite.go
alovak: Documentation for `Composit.ReadFrom` should be updated. ReadFrom will always return bytes read.

I missed that ReadFrom reads until the EOF.

adamdecaf:
I think Pack/Unpack can be implemented with ReadFrom and WriteTo. Either way backwords compatibility would be nice.

field/composite.go
krishishah: Could we also update the `Field` and `Encoder` interfaces to indicate that the return type in `ReadFrom` is always measured in bytes as well?

encoding/bcd.go
alovak: Adam, I named it so. I don't see how we can separately use Encoder and Decoder. They always go in pairs and in the `Spec` you specify one thing that will be both used for encoding and decoding.

I just named it `Coder` to avoid confusion.

alovak:
This PR is ready. I returned `Pack` and `Unpack` to the message. Internally they use `ReadFrom` and `WriteTo`.

alovak:
@adamdecaf I implemented Writer:
```go
func (m *Message) Write(p []byte) (n int, err error) {
	n, err = m.ReadFrom(bytes.NewReader(p))
	return
}
```

but I'm not sure that Writer in its form will work properly. `ReadFrom` reads from the stream while it's decoding field by field. It controls how much data we need to read. Stream is responsible to provide the data.

For `Writer` interface, I don't understand how we can write `p` into `m` and iterate over fields when there is not enough data for the next field to read? When should we initiate parsing of the written data?

Any ideas or advice on this `Write`?

While Read method totally makes sense:

```go
func (m *Message) Read(p []byte) (n int, err error) {
	if m.packed == nil {
		m.packed, err = m.Pack()
		if err != nil {
			return 0, err
		}

	}
	if m.packedReadIndex >= len(m.packed) {
		err = io.EOF
		m.packed = nil
		return
	}

	n = copy(p, m.packed[m.packedReadIndex:])
	m.packedReadIndex += n
	return
}
```

adamdecaf:
Did we come to a conclusion on this?

alovak:
Closing this for now as benefits from such huge change is not obvious.

PR #80, opened 2021-07-12
add CLI with describe

alovak:
This PR adds CLI that displays ISO8583 messages in human-readable format.

Examples:
```
➜ ./bin/iso8583 describe msg.dat
ISO 8583 Message:
MTI...........................: 0100
Bitmap........................: 7000000000000000
Bitmap bits...................: 01110000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
F002 Primary Account Number...: 4242424242424242
F003 Processing Code..........: 123456
F004 Transaction Amount.......: 100
```

```
➜ ./bin/iso8583 describe
Display ISO 8583 file in a human-readable format.

Usage:
  iso8583 describe <files> [flags]

Flags:
  -spec string
        name of built-in spec (default "87")
```

What's not included in this PR (will be added later):
* it works only with one pre-defined ISO spec (spec87)
* no pre-built binary CLI


describe/describe.go
```diff
@@ -0,0 +1,94 @@
+package describe
```
alovak, line 1: not sure, but maybe I can just copy these all into cmd/describe...

wadearnold:
This looks great. A cut and paste of the entire bytes in string that was passed until unpack would be nice as well.

wadearnold (approved):

wadearnold:
so awesome!

cmd/iso8583/describe.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/moov-io/iso8583"
+	"github.com/moov-io/iso8583/describe"
+)
+
+var specs = map[string]*iso8583.MessageSpec{
+	"87": iso8583.Spec87,
```
adamdecaf, line 13: Is this here so we could have `-spec 91` to change which one is used?

cmd/iso8583/main.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+var (
+	programName = filepath.Base(os.Args[0])
```
adamdecaf, line 11: I think `flag.Arg(0)` would be safer. It'll strip out the CLI arguments.

```
$ go doc flag arg
package flag // import "flag"

func Arg(i int) string
    Arg returns the i'th command-line argument. Arg(0) is the first remaining
    argument after flags have been processed. Arg returns an empty string if the
    requested element does not exist.

```

cmd/iso8583/main.go
```diff
@@ -0,0 +1,69 @@
...
+		fmt.Fprintf(os.Stdout, "Work seamlessly with ISO 8583 from the command line.\n\nUsage:\n  %s <command> [flags]\n\n", programName)
+		fmt.Fprintf(os.Stdout, "Available commands:\n")
+
+		// TODO: we will print all commands when we have more than one
+		fmt.Fprintf(os.Stdout, "  %s: display ISO 8583 file in a human-readable format\n", describeCmd)
+		fmt.Fprintf(os.Stdout, "\n")
+	}
+
+	describeCommand.Usage = func() {
+		fmt.Fprintf(os.Stdout, "Display ISO 8583 file in a human-readable format.\n\nUsage:\n  %s %s <files> [flags]\n\n", programName, describeCmd)
```
adamdecaf, line 27: This is a nitpick, but I've found Go's backtick strings are handy here. No need to change - just mentioning. 

```suggestion
		fmt.Fprintf(os.Stdout, `Display ISO 8583 file in a human-readable format.
		
Usage:
  %s %s <files> [flags]
`, programName, describeCmd)
```

cmd/iso8583/main.go
```diff
@@ -0,0 +1,69 @@
...
+
+	switch os.Args[1] {
+	case describeCmd:
+		describeArgs := os.Args[2:]
+		if len(describeArgs) == 0 {
+			describeCommand.Usage()
+			os.Exit(1)
+		}
+
+		describeCommand.Parse(os.Args[2:])
```
adamdecaf, line 51:
```suggestion
	case describeCmd && len(os.Args) > 1:
		describeArgs := os.Args[2:]
		if len(describeArgs) == 0 {
			describeCommand.Usage()
			os.Exit(1)
		}

		describeCommand.Parse(describeArgs)
```

describe/describe.go
adamdecaf: I did that because I knew we'd import the human readable format into other projects. I knew ach-web-viewer was something I wanted to created.

describe/describe.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (p *fieldPrinter) addField(description, value string) {
+	field := printableField{description, value}
+	if len(description) > p.maxLen {
+		p.maxLen = len(description)
+	}
+
+	p.fields = append(p.fields, field)
+}
```
adamdecaf: `text/tabwriter` is super helpful here. You might want to use it.

https://pkg.go.dev/text/tabwriter

cmd/iso8583/describe.go
alovak: right. Hope that we can add more specs :)
I'm also going to add `--spec-file xxx.json` to load spec from JSON.

cmd/iso8583/main.go
alovak: I checked `flag.Arg(0)`

* It returns empty string if `flag.Parse()` was not called. We want to call `flag.Parse()` and use program name in `flag.Usage()` on error.
* According to the documentation, it returns first argument (command) when `flag.Parse()` was called:

```
➜ ./bin/iso8583 hello
flag.Arg(0) = hello
```

describe/describe.go
alovak: Thanks @adamdecaf! Initially, I didn't see how I can use it because I didn't want the second column to be aligned. Then I read the docs and examples and found that I just have to omit the trailing tab for the trailing text.

It's more compact now :)

PR #96, opened 2021-07-16
Composite field: stop parsing bytes once length indicated by prefix is reached

krishishah:
Bug fix: If unpacking fields by ID, the composite field type currently does not stop parsing bytes when the length indicated by the prefix is reached.

This PR ensures that unpacking returns when that limit is reached.


alovak (approved):
Looks good! Thanks!

PR #110, opened 2021-07-22
JSON export/import of Spec and CLI describe with json spec file

mfdeveloper508:


cmd/iso8583/describe.go
```diff
@@ -41,6 +41,33 @@ func Describe(paths []string, specName string) error {
 	return nil
 }
 
+func DescribeWithSpecFile(paths []string, specFileName string) error {
```
alovak, line 44: I would extract core functionality into `func describe(paths, spec) error` and then inside `DescribeWithSpecFile` and `Describe` call this `describe(path, spec)` function.

specs/builder.go
```diff
@@ -0,0 +1,312 @@
...
+
+	return outputBuffer.Bytes(), nil
+}
+
+func getPrefixInterface(prefixName string, length int) prefix.Prefixer {
+
+	prefixName = "*" + prefixName
+	var pref prefix.Prefixer
+
+	switch prefixName {
```
alovak, line 201: technically, you can just create a map:

```
var prefixes = map[string]prefix.Prefixer{
  "ASCII.Fixed": prefix.ASCII.Fixed,
  "ASCII.L": prefix.ASCII.L,
  "ASCII.LL": prefix.ASCII.LL,
  ...
}
```

then we don't need to care about prefix lenght.

specs/builder.go
```diff
@@ -0,0 +1,312 @@
...
+		} else if length == 3 {
+			pref = prefix.EBCDIC.LLL
+		} else if length == 4 {
+			pref = prefix.EBCDIC.LLLL
+		}
+	}
+
+	return pref
+}
+func getEncodingInterface(encName string) encoding.Encoder {
```
alovak, line 279: we can use a map to get encoding as well.

Or (crazy idea), we can try to create encoding using reflect package if we know its name. Kind of: `enc := eval("package.${encName}")` (not sure how to do it).

spec87ascii.json
```diff
@@ -0,0 +1,549 @@
+{
+	"name": "ISO 8583 v1987 ASCII",
+	"fields": {
+		"Field000": {
```
alovak, line 4: I think that it would be great if we can follow go spec as close as possible. Here, instead of "Field000", I would jsut use it as "0", "1", "2", ... "32". Without "Field" prefix.

alovak:
@mfdeveloper508 I really like your approach with using builder to make/load JSON spec!
Great job 💪! Let's decide on some aspects and it's good to go!

spec87ascii.json
mfdeveloper508: sure, we can use like this. But the naming isn't fit for XML format. 
xml tag as number is difficult to use

cmd/iso8583/describe.go
mfdeveloper508: fixed

specs/builder.go
mfdeveloper508: fixed

specs/builder.go
mfdeveloper508: it seems that the approach isn't fit for var types with length. So used resource map

spec87ascii.json
alovak: Hm... I didn't thought about XML. Maybe we can keep it without Field in JSON and use field for XML only? I don't know how many XML users we are going to have...

specs/builder.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to export message spec")
+	}
+
+	return outputBuffer.Bytes(), nil
+}
+
+func getPrefixInterface(prefixName string, length int) prefix.Prefixer {
+
+	var prefixes = map[string]prefix.Prefixer{
```
alovak, line 199: to not re-define it on every call of the function, we can move the var definition outside of the func.

specs/builder.go
```diff
@@ -0,0 +1,256 @@
...
+	err := enc.Encode(dummyJson)
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to export message spec")
+	}
+
+	return outputBuffer.Bytes(), nil
+}
+
+func getPrefixInterface(prefixName string, length int) prefix.Prefixer {
```
alovak, line 197: we don't need the `length`, as it's already in `L`, `LL`, `LLL` and is mapped to the prefix of a specific length.

specs/builder.go
```diff
@@ -0,0 +1,256 @@
...
+	if length > 0 {
+		prefixName = prefixName + "." + strings.Repeat("L", length)
+	}
+
+	return prefixes[prefixName]
+}
+
+func getEncodingInterface(encName string) encoding.Encoder {
+
+	var encodes = map[string]encoding.Encoder{
```
alovak, line 232:
```suggestion
	var encodings = map[string]encoding.Encoder{
```

specs/builder.go
alovak: let's move `var encodings` out of the `func`

cmd/iso8583/describe.go
```diff
@@ -6,7 +6,7 @@ import (
 	"os"
 
 	"github.com/moov-io/iso8583"
-	"github.com/moov-io/iso8583/cmd/iso8583/describe"
+	msg "github.com/moov-io/iso8583/cmd/iso8583/describe"
```
alovak, line 9: Lets rename`func describe` into `func describeMessage` and keep package as `describe`.

It's not reading well `msg.Message`, while `describe.Message` is.

specs/builder.go
```diff
@@ -0,0 +1,256 @@
...
+
+		enc := getEncodingInterface(dummyField.Enc)
+		pref := getPrefixInterface(dummyField.Prefix, dummyField.PrefixLength)
+		pad := getPadInterface(dummyField.Padding)
+
+		var newField field.Field
+
+		switch dummyField.Type {
+		case reflect.TypeOf(field.String{}).Name():
+			newField = field.NewString(&field.Spec{
```
alovak, line 83: I suggest here two things:
1. let's define spec before `switch` like this

```go
spec: = &field.Spec{
		Length:      dummyField.Length,
		IDLength:    dummyField.IDLength,
		Description: dummyField.Description,
		Enc:         enc,
		Pref:        pref,
		Pad:         pad,
	}
```

and then, let's create a map for the field constructor, like this:

```
var fieldsConstructor = map[string]func(spec *field.Spec) *field.Field {
  "String": func(spec *field.Spec) { return field.NewString(spec) },
  "Numeric": func(spec *field.Spec) { return field.NewNumeric(spec) },
  // ...
}
```

then inside we can do:

```go
fieldConstructor := fieldsConstructor[dummyField.Type]
if fieldConstructor == nil {
  // handle error
}

newField := fieldConstructor(spec)

// ...
```

mfdeveloper508: fixed

specs/builder.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/moov-io/iso8583/padding"
+	"github.com/moov-io/iso8583/prefix"
+)
+
+const (
+	FieldPrefix = "Field"
+)
+
+var (
+	fieldConstructor = map[string]func(spec *field.Spec) field.Field{
```
alovak, line 23: @mfdeveloper508 sorry, but I forgot that we can use function as a type and change the map into this:

```go
type fieldConstructorFunc func(spec *field.Spec) field.Field

fieldConstructor = map[string]fieldConstructorFunc{
"String": field.NewString,
"Numeric": field.NewNumeric,
// ...
}
```

specs/builder.go
```diff
@@ -0,0 +1,216 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid field index, index's format is `Field`+index ")
+		}
+
+		constructor := fieldConstructor[dummyField.Type]
+		if constructor == nil {
+			return nil, fmt.Errorf("unable create field with %s", dummyField.Type)
+		}
+
+		enc := encodings[dummyField.Enc]
```
alovak, line 119: It would be great to check that value was found and if not, return the error message. We want developers know that encoding (prefixer, padder, etc.) with the provided name was not found.

PR #113, opened 2021-07-23
remove filed method to message added

shamigor:


alovak (approved):
Thanks @shamigor! This looks great!

May I ask you to share what is your use case for this change? It may help us (and community) better understand why and how it can be useful

shamigor:
> Thanks @shamigor! This looks great!
> 
> May I ask you to share what is your use case for this change? It may help us (and community) better understand why and how it can be useful

shamigor:
> Thanks @shamigor! This looks great!
> 
> May I ask you to share what is your use case for this change? It may help us (and community) better understand why and how it can be useful

Yes, of course

We have to answer our provider with the same message as the request message ( but exclude some fields and add some new fields).
For example, in the request message, we have 10 fields, in the response message we have the same 9 fields and a new one

I don't know why is that. I'll ask about it our provider

vtolstov:
Yes, some answers contains request fields. But you don't right. You need to construct new messages with needed fields.

shamigor:
> Yes, some answers contains request fields. But you don't right. You need to construct new messages with needed fields.

In our case, we have requests and responses with only one different field. 
We see that we have opportunities to change MTI or Field value (message is mutable), why do we cannot remove a field from the message?
And what problems will we have if we'll use the same message in response? 

Thanks a lot for answering

vtolstov:
I think that after parsing we must forbid message fields changing from user publuc funcs.
You case can bring errors, for ex other side add additional field that must not present in response

alovak:
@vtolstov I agree with you that modifying the original message is not a good idea. At the same time, we need to find a way how to create a message copy where you can replace fields/values and it will be safe.

vtolstov:
Im prefer to have Copy(src message, []fields_num) NewMessage

alovak:
@shamigor I think that we can mutate the iso8583 message. What you suggest is good (for now :). We either have to list fields to copy (which is not really convenient) or remove unnecessary fields from the copied message.

I see one technical issue with this PR. When we remove the field from the `fieldsMap`only, we still have it inside `fields`. When we call `Pack` method, we iterate over `fields` and if typed API was used to set a field value, we will restore it inside `fieldsMap` (basically we revert the removal of the field). Here is the code I'm talking about:

```
func (m *Message) Pack() ([]byte, error) {
    // ...
    ids, err := m.setPackableDataFields()
    // ...
}

func (m *Message) setPackableDataFields() ([]int, error) {
	// ...

	for id, field := range m.fields {
		// ...

		// These fields are set using the typed API
		if m.dataValue != nil {
			dataField := m.dataFieldValue(id)
			// no non-nil data field was found with this name
			if dataField == (reflect.Value{}) || dataField.IsNil() {
				continue
			}
			if err := field.SetData(dataField.Interface()); err != nil {
				return nil, fmt.Errorf("failed to set data for field %d: %w", id, err)
			}

			// mark field as set
			m.fieldsMap[id] = struct{}{}
		}

		// ...

	}
	//...

}
```

So, we should think about how to remove the field from `m.dataValue` as well.

Or, find a convenient way to make a message copy without some fields.

shamigor:
@alovak I've added a fix for typed fields, thanks for the feedback

alovak:
@shamigor thanks for the update! It looks good! Let's wait for a little bit on the decision on how to make a copy of the message (#118 )

alovak:
Is it still relevant?

Dakinola892 (approved):
Having a way to bulk remove fields would be useful for constructing response messages, in comparison with .Clone() - I think this is an impactful change

adamdecaf:
Is this issue still relevant? Have we come to a conclusion on what to do (if anything)?

alovak:
It seems that no. I'll keep it open for a little bit.

alovak:
Based on my experience, the best way is to define message struct without the field that you want to remove, unmarshal message values into the struct, then create a new message and marshal the struct into the new message.

The first option (error handling is not here :):

```go

	// define struct for the response
	type authorizationResponse struct {
		F0 *field.String
		F1 *field.Bitmap
		F2 *field.String
		// let's say we don't have F3 in the response
		F4 *field.String
	}

	// handle request
	requestMessage := iso8583.NewMessage(spec)
	err := requestMessage.Unpack(rawMessage)

	// get all data except fields you don't need
	responseData := &authorizationResponse{}
	err = requestMessage.Unmarshal(responseData)

	// send only needed data back
	responseMessage := iso8583.NewMessage(spec)
	err = responseMessage.Marshal(responseData)
	rawMessage, err := responseMessage.Pack()

```

the alternative option is to define one struct and share it between request and response. You just have to set the field into nil, so the message field will not be populated:

```go
	// define a struct for the request
	type authorizationRequest struct {
		F0 *field.String
		F1 *field.Bitmap
		F2 *field.String
		F3 *field.String
		F4 *field.String
	}

	// handle request
	requestMessage := iso8583.NewMessage(spec)
	err := requestMessage.Unpack(rawMessage)

	// get all data into struct
	requestData := &authorizationRequest{}
	err = requestMessage.Unmarshal(requestData)

	// get same data into different instance of struct
	responseData := &authorizationRequest{}
	err = requestMessage.Unmarshal(requestData)
	// remove fields by setting them to nil
	responseData.F3 = nil

	responseMessage := iso8583.NewMessage(spec)
	err = responseMessage.Marshal(responseData)
	rawMessage, err := responseMessage.Pack()
```

Please, let me know if it makes sense.

alovak:
@shamigor I'm closing the PR. Please, let me know if the solution(s) described above help.

alovak:
Here is an example of how you may want to build response (based on the conversation in our community Slack):

```go
	// received request
	request := &NetworkManagementRequest{}
	err := requestMessage.Unmarshal(request)
	if err != nil {
		return fmt.Errorf("unmarshaling request message: %w", err)
	}

	response := &NetworkManagementResponse{
		// set your MTI
		MTI:                   field.NewStringValue(MessageTypeIndicatorNetworkManagementResponse),
		TransmissionDateTime:  request.TransmissionDateTime,
		STAN:                  request.STAN,
		NetworkManagementCode: request.NetworkManagementCode,
		// set your response code
		ResponseCode:          field.NewStringValue(ResponseCodeApproved),
	}
	responseMessage := iso8583.NewMessage(ISO8583MessageSpec)
	err = responseMessage.Marshal(response)
	if err != nil {
		return fmt.Errorf("marshaling response message: %w", err)
	}
```

PR #115, opened 2021-07-28
Pack and read MTI using typed API

krishishah:
Currently, the MTI can neither be read nor set using the typed API. This PR implements behaviour to allow for that to happen.


wadearnold (approved):

alovak (approved):
Thanks!!!

PR #121, opened 2021-08-25
Add support for the encoding and decoding of BER-TLV fields 

krishishah:
Unfortunately, this PR is much larger than desired and involves breaking changes. However, these breaking changes are necessary for this library to decode and encode TLVs.

## New Behaviour

As of this PR, users are now able to define `BER-TLV` (and SIMPLE-TLV fields if they'd like, although additional encoders and prefixers would need to be defined to support SIMPLE-TLV) fields within their ISO8583 specs as such:

```
spec := &MessageSpec{
       ...
	// TLV
	55: field.NewComposite(&field.Spec{
		Length:      999,
		Description: "ICC Data – EMV Having Multiple Tags",
		Pref:        prefix.ASCII.LLL,
		Tag: &field.TagSpec{
			Enc: encoding.BerTLVTag,
		},
		Subfields: map[string]field.Field{
			"9A": field.NewString(&field.Spec{
				Description: "Transaction Date",
				Enc:         encoding.Binary,
				Pref:        prefix.BerTLV,
			}),
			"9F02": field.NewString(&field.Spec{
				Description: "Amount, Authorized (Numeric)",
				Enc:         encoding.Binary,
				Pref:        prefix.BerTLV,
			}),
		},
	}),
	...
}
```

Along with this, they're able to define the following `Data` structs to match the spec:

```
type F55Data struct {
	F9A   *String
	F9F02 *String
}
```

And encode and decode data in and out of this struct given the message spec has been defined as above. Because BER-TLV keys fields by hex rather than int (as data elements and subelements do), we are required to make breaking changes to support this field encoding.

## Breaking Changes

The following breaking changes have been made:

### field.Spec

We have replaced the `IDLength` field of the `field.Spec` struct with a more complete `*TagSpec` sub-struct:

```
// TagSpec is used to define the format of field tags (sometimes defined as field IDs).
// This is most commonly used by composite field types such as TLVs, subfields
// and subelements. TagSpecs need not be defined for primitive field types
// such as numerics, strings or for composite field types that don't consist
// of tags in the message payload.
type TagSpec struct {
	// Length is defined for subfields and subelements whose tag
	// lengths are fixed and can be defined statically.
	// This field should not be populated in conjunction with the TLV Tag
	// encoder as their lengths are determined dynamically.
	Length int
	// Enc defines the encoder used to marshal and unmarshal
	// the tag.
	Enc encoding.Encoder
	// Pad sets the padding direction and type of the tag.
	// This is most commonly used for composite field types
	// whose tags hold leading 0s e.g. '003' would be unpadded to '3'.
	Pad padding.Padder
}
```

We have also used this opportunity to replace `Fields map[int]Field` with `Subfields[string]Field`. Together these changes look like:

```
// Spec defines the structure of a field.
type Spec struct {
...
	// Tag sets the tag specification. Only applicable to composite field
	// types.
	Tag *TagSpec
        // Subfields defines the subfields held within the field. Only
	// applicable to composite field types.
	Subfields map[string]Field
...
}
```

This allows us to define custom padding and encoding configuration for the `Tag` (formerly `ID`) component of a field whilst also being able to 'key' subfields in formats outside of just ints e.g. hex for TLV. The `field.Composite` type takes advantage of this additional configuration to now support `Subfields`, `Subelements` and now `BER-TLV` format fields.

We felt that from a semantic POV, renaming `Spec.Fields` to `Spec.Subfields` made the purpose of the field much clearer for users of the library.

The move from `ID` to `Tag` hasn't changed the semantics of the library - the two are interchangeable. However, since TLVs are a more general purpose composite field standard (with its own ISO standard), we felt that it would make the API more self-documenting to refer to `Tags` rather than `IDs`.

More can be seen on how this looks in the unit tests of this PR.

### field.OrderedMap

Because we now key subfields by `string` rather than `int`,`OrderedMap` is now of type `map[string]Field`. This hasn't made a huge difference to the implementation and the actual JSON that is returned is unchanged.

### Prefixer

`DecodeLength` of the `prefix.Prefixer` interface now returns an additional `read int` argument as such: `DecodeLength(maxLen int, data []byte) (length int, read int, err error)`. This is a direct replacement for the `Length() int` function which has now been removed. The previous design of the interface assumed that field prefixes are defined statically e.g. `Fixed`, `LVar`, `LLVar`, `LLLVar`. However, the `Length` component of BER-TLV fields are encoded and decoded dynamically. 
 
It seemed awkward to have to make the `BerTLV` prefixer behave in a stateful way where it stores the length in memory everytime a field length has been decoded. Therefore, we have opted to remove it.

Regardless, this change to the API is an improvement as `Length()` has always been called immediately after `DecodeLength()` to calculate offsets for the next field operation. As such, we've made it less chatty whilst also supporting additional use cases.

### Encoding

When implementing TLV hex encoding / decoding, we found the need to do Hex-ASCII encoding in the opposite direction to that already implemented.

As such, we've renamed the existing `encoding.Hex` encoder to `encoding.HexToASCII` (meaning it encodes hex to ascii). In addition, we have introduced a separate `encoding.ASCIIToHex` encoder to perform encoding in the opposite direction.

Before doing this, I flirted with the idea of just using the existing encoder in the opposite direction. However, this wouldn't work for TLV field types e.g. `field.String` that call `Encode` and `Decode` within `Pack` and `Unpack` already.  

## Sources

The following sources were used to implement this:

* [ISO 7816-4 TLV](https://github.com/dongri/emv-qrcode-doc/blob/master/ISO%20IEC%207816-4.pdf) 
* VISA BASE I Technical Specifications, Volume 1
* [EMV TLV Decoder](https://emvlab.org/) - All of our test cases may be decoded using this tool
* [EMVCo Book 3](https://www.emvco.com/wp-content/uploads/2017/05/EMV_v4.3_Book_3_Application_Specification_20120607062110791.pdf) - Annex B Rules for BER-TLV Data Objects
 * [Blog: Encoding Integers in the EMV Protocol](https://www.lightbluetouchpaper.org/2010/01/19/encoding-integers-in-the-emv-protocol/)


wadearnold (approved):

field/ordered_map.go
```diff
@@ -8,15 +8,15 @@ import (
 )
 
 // Custom type to sort keys in resulting JSON
-type OrderedMap map[int]Field
+type OrderedMap map[string]Field
```
alovak, line 11: @krishishah with the change, we don't get proper order in JSON for numeric IDs. Here is an example of what we may get (when we have 17 and 107 as IDs):

```json
{
  "0":"0100",
  "1":"D00080000000000000000000002000000000000000000000",
  "107":"102",
  "17":"101",
  "2":"4242424242424242",
  "4":"100"
}
```

I think that we may keep OrderedMap as is for numeric IDs and just use default JSON marshalling for a composite field without sorting. Or, we may add `OrderedStringMap` in addition to `Ordered(Int)Map`. What do you think?

krishishah: Great spot @alovak !

I don't think we should should maintain two different implementations of `OrderedMap` (i.e. `OrderedIntMap` and `OrderedStringMap`).

Ideally, we'd just generify `OrderedMap` to accommodate for different sorting requirements.

Thankfully, the `sort` pkg gives us much of the behaviour we need.

The way I see it is we have two options:

1. We can change the `OrderedMap` type to:

```Go
// Custom type to sort keys in resulting JSON                                                                                                                                                           
type OrderedMap struct {                                                                                                                                                                                
    Fields map[string]Field                                                                                                                                                                             
    Sort sort.StringSlice // "github.com/moov-io/iso8583/sort"                                                                                                                                                                          
}
```

The big change here would be to call `om.Sort(keys)` instead of `strings.Sort(keys)`. Our two big users of this type are `Message` and `field.Composite`. Each of them can quite easily pass in the correct `sort` funcs to determine the ordering of those keys.

One optimisation we can make is to push the sorting of these keys onto the callers of `OrderedMap`. This is more efficient to do for `fields.Composite` because we already maintain an ordered list of fields and this leads us onto the next option.

2. Instead of holding a `map[string]Field`, `OrderedMap` could now become:

```go
type KeyVal struct {
    Key string
    Val Field \\ this could even be an interface{} if we want to generify it further.
}

// Define an ordered map
type OrderedMap []KeyVal
```

Users of this field can populate `[]KeyVal` in an order of their choice and delegate marshalling to `OrderedMap.MarshalJSON()`

What are your thoughts?

PR #123, opened 2021-09-02
Add support for configurable subfield sorting for composite fields

krishishah:
Currently, the `Composite` field type performs basic but crude string sorting using the following function:

```
func orderedKeys(kvs map[string]Field) []string {
	keys := make([]string, 0)
	for k := range kvs {
		keys = append(keys, k)
	}
	sort.Strings(keys)
	return keys
}
```

This worked well before when we keyed subfields by `int`. However, our `subfields` now leverage `strings` to either key by `int` (for subfields and subelements) or `hex` (for TLVs). Relying on the function above alone means that calling `orderedKeys` on `[]string{"1", "2", "11"}` would mutate the slice to `[]string{"1", "11", "2"}`.

This PR introduces a new `sort` package with a `type StringSlice func(x []string)` function type.

This function type is referenced in the `TagSpec` struct defined in `field/spec.go` and is used to order subfields on initialisation of a `Composite` field type. In fact, the `Sort` parameter is now a required field in the `TagSpec` of a `Composite` field type. As such, this PR could be construed to be breaking because of this. I'll leave this detail up to the discretion of the reviewer

### DIFF
--- a/encoding/bertlv.go
+++ b/encoding/bertlv.go
@@ -26,8 +26,8 @@ func (berTLVEncoderTag) Encode(data []byte) ([]byte, error) {
 // 2) We must continue reading subsequent bytes until we arrive at one whose
 //    most significant bit is unset.
 //
-// On success, the ASCII representation of the Tag as well are returned along
-// with the number of bytes read e.g. []byte{0x5F, 0x2A} would be converted to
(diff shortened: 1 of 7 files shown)


wadearnold (approved):
Great use of the sort interface

adamdecaf (approved):

PR #124, opened 2021-09-08
utility track data

mfdeveloper508:


mfdeveloper508:
Track 3 has several elements in security data and additional data
but track 3 has been implemented easily using dummy element
Should we implement track 3 with detailed data format?

wadearnold:
@mfdeveloper508  let's pull out the track 3 secure element data into its own issue and address it separately. What you have implemented will work for a magnetic card swipe.

wadearnold (approved):
@alovak can you merge this after your review and cut a release with a version and release notes.

track/regen.go
```diff
@@ -0,0 +1,239 @@
...
+	return nil
+}
+
+// Generator generates random strings.
+type Generator interface {
+	Generate() string
+	String() string
+}
+
+func NewGenerator(pattern string, inputArgs *GeneratorArgs) (generator Generator, err error) {
```
krishishah, line 34: Any reason we're publicly exposing this func as well as the `Generator` interface? I don't

alovak:
I'm on it.

alovak:
@mfdeveloper508 thanks for a great job! I see 

I want to suggest some changes before we merge it with the master.

It would be great to implement track1, track2 (and maybe track3) as a `field.Field`, not just as a utility class. If track implements the `field.Field` interface, we can easily use it in our specifications. We can also use them separately as utility class. Here is what I mean:

```go
// Example 1, create track 1 field using
track1 := track.NewTrack1Value(&track.Track1Data{
	Number:      "",
	Name:        "",
	ServiceCode: "",
	// ...
})

raw, err := track1.Pack()
// handle error

track1 := field.NewTrack1(&field.Spec{
	Length:      76,
	Description: "Track 1 Data",
	Enc:         encoding.ASCII,
	Pref:        prefix.ASCII.LL,
})

read, err := track1.Unpack("011234567890123445=724724000000000****00300XXXX020200099010=********************==1=100000000000000000**")
// handle error

trackData := track1.Data().(*track.Track1Data)
trackData.ServiceCode
trackData.Number
trackData.Name
trackData.CVV
trackData.DiscretionaryData
// ....

```

Changes that are required:
* move `./track` into `./field` -> `/field/track/track1`
* implement `Field` interface for `Track1` and `Track2`
* add `Data()` method so we can get value out of the track field

Thoughts, comments?

track/track.go
```diff
@@ -0,0 +1,138 @@
...
+
+type GeneralCard struct {
+	// Available for Track 1, Track 3
+	FormatCode string `json:"format_code"`
+
+	// Available for Track 1, Track 2, Track 3
+	PrimaryAccountNumber string `json:"primary_account_number"`
+
+	// Available for Track 1, Track 2, Track 3
+	CardType string `json:"card_type"`
```
alovak, line 42: While I see the benefit of detecting card type, I think that current model should be changed a little bit into something like:

* Card
  * Track 1
  * Track 2

It means that `Card` contains tracks and then we can add `Type` into `Card` object.

If there is no specific need for CardType here, let's remove it.

track/track.go
```diff
@@ -0,0 +1,138 @@
...
+	// Available for Track 1, Track 2
+	ServiceCode string `json:"service_code"`
+
+	// Available for Track 1, Track 2, Track 3
+	//  If track 3, the field describe security data + additional data
+	DiscretionaryData string `json:"discretionary_data"`
+}
+
+// Track is network header interface to write/read encoded message legnth
+type Track interface {
```
alovak, line 59: As I suggested in my first comment to the PR, let's implement `field.Field` interface in `Track1` and `Track2`.

track/track.go
```diff
@@ -0,0 +1,138 @@
...
+	if len(str) == 0 {
+		return nil, nil //skip
+	}
+
+	if len(str) == 0 {
+		return nil, errors.New("invalid expired date string")
+	}
+
+	str = str[:2] + "/" + str[2:4]
+	r, err := regexp.Compile(expiryDatePattern)
```
alovak, line 117: I would use `regexp.MustCompile(...)` at the package level (outside of the function) to compile once and just use compiled regexp in the function then.

krishishah (commented):
I agree with @alovak 's comments on the whole. It feels more intuitive to implement much of this within the `field` package.

I'd also rename `First`, `Second`, `Third` as `Track1`, `Track2`, `Track3` etc

I personally don't see the value of the `Generator` pattern myself and see lots of opportunity to simplify the code in its current state. However, it's possible that I don't fully appreciate what its there for.

As a general pointer, could we also document all of our public structs, fields, functions and methods and add examples wherever possible? It makes it much easier to follow the code.

track/track.go
```diff
@@ -0,0 +1,138 @@
...
+message.Field(3, "123456")
+message.Field(4, "100")
+message.Field(4, "100")
+
+dataBuf, _ := NewTrackSecond().Write(card)
+message.Field(36, string(dataBuf))
+
+*/
+
+type GeneralCard struct {
```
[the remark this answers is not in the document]
krishishah: If we're moving this to be within `field/track` then we could even just call it `Data`?

No need to repeat `track.TrackData`.

But :100: to the rename from `GeneralCard` to something more concise

krishishah:
> move ./track into ./field -> /field/track/track1

@alovak is there a specific reason you don't just want to have this logic live inside of `field/`? Do we need the nested `field/track` pkg?

track/track.go
```diff
@@ -0,0 +1,138 @@
...
+	expired := ExpiryDate(time.Date(2000+year, time.Month(month), 1, 0, 0, 0, 0, time.UTC))
+
+	return &expired, nil
+}
+
+func (t ExpiryDate) String() string {
+	if (time.Time)(t).IsZero() {
+		return ""
+	}
+	return fmt.Sprintf("%2s%02d", strconv.Itoa((time.Time)(t).Year())[2:4], (time.Time)(t).Month())
```
alovak, line 137: should it include `/`?

track/track.go
```diff
@@ -0,0 +1,138 @@
...
+	expiryDatePattern    = `^([0-9]{2})\/?(0[1-9]|1[0-2])$` //YYMM
+	trackFirstPattern    = `^([A-Z]{1})([0-9]{1,19})\^([^\^]{2,26})\^([0-9]{4}|\^)([0-9]{3}|\^)([^\?]+)$`
+	trackSecondPattern   = `^([0-9]{1,19})\=([0-9]{4}|\=)([0-9]{3}|\=)([^\?]+)$`
+	trackThirdPattern    = `^([0-9]{2})([0-9]{1,19})\=([^\?]+)$`
+	trackFirstMaxLength  = 76
+	trackSecondMaxLength = 37
+	trackThirdMaxLength  = 104
+)
+
+type ExpiryDate time.Time
```
alovak, line 105: Not sure I see the benefit from introduced `ExpiryDate` type. If we use `time.Time` in `GeneralCard` (or in `TrackData` when we change it) we just need two functions:
1. to parse/return `time.Time`
2. to return string representation like `t.Format("06/01")`

track/track_test.go
```diff
@@ -0,0 +1,76 @@
...
+	t.Run("Track 1 data", func(t *testing.T) {
+		tracker := NewTrackFirst(true)
+
+		samples := []string{
+			`B4815881002867896^YATES/EUGENE JOHN         ^21129821000123456789`,
+			`B4815881002861896^YATES/EUGENE L            ^^^356858      00998000000`,
+			`B4000340000000506^John/Doe                  ^13011110000123000`,
+			`B1234567890123445^PADILLA/L.                ^99011200000000000000**XXX******`,
+		}
+		for _, sample := range samples {
```
alovak, line 20: It would be great if we test parsed value here as well. From my experience, it's not enough to read/write without error. We should check what exactly we get in `Number`, `Expiration` and other fields.

track/track_test.go
```diff
@@ -0,0 +1,76 @@
...
+			require.Equal(t, sample, string(buf))
+		}
+	})
+	t.Run("Track 1 data with invalid name length", func(t *testing.T) {
+		tracker := NewTrackFirst(false)
+
+		samples := []string{
+			`B4242424242424242^SMITH JOHN Q^11052011000000000000`,
+		}
+		for _, sample := range samples {
```
alovak, line 34: From the name of the test, I would expect us to test for some error here. But I don't see it. What's the expected behavior? Not to fail (ignore name length?)?

track/track_test.go
```diff
@@ -0,0 +1,76 @@
...
+	t.Run("Track 2 data", func(t *testing.T) {
+		tracker := NewTrackSecond()
+
+		samples := []string{
+			`4000340000000506=2512111123400001230`,
+			`4242424242424242=15052011000000000000`,
+			`1234567890123445=99011200XXXX00000000`,
+		}
+		for _, sample := range samples {
+			card, err := tracker.Read([]byte(sample))
```
alovak, line 52: As in my previous comment, we need to check if we parse values properly. As with table tests I would use `input` and `want` structure with expected values.

alovak:
> > move ./track into ./field -> /field/track/track1
> 
> @alovak is there a specific reason you don't just want to have this logic live inside of `field/`? Do we need the nested `field/track` pkg?

We can put it just into field. Initially I thought that we will need more files to support `track` field. But it seems we can put it all into one file.

track/regen.go
```diff
@@ -0,0 +1,239 @@
...
+	var gen *internalGenerator
+	gen, err = newGenerator(regexp, &args)
+	if err != nil {
+		return
+	}
+
+	return gen, nil
+}
+
+func init() {
```
alovak, line 60: Can we do the same without `init()`?

track/track.go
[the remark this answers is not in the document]
alovak: It seems we don't even need `Data` as `Track1` is already that field. At least this is how it works for `String`, `Number` and the rest.

track/regen.go
```diff
@@ -0,0 +1,239 @@
+package track
```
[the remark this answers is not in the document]
mfdeveloper508: that's right, 
'reqen' is short for regenerator of regexp
main feature is to create string with regex pattern and params

field/track.go
```diff
@@ -0,0 +1,404 @@
...
+		return true
+	}
+	return false
+}
+
+const (
+	VersionFirst  = 1
+	VersionSecond = 2
+	VersionThird  = 3
+)
```
louisheath, line 59: If these are to be of type `TrackVersion` lets define them so, as currently they are `int`s
```suggestion
type TrackNumber int

const (
    Track1 TrackNumber = 1
    Track2 TrackNumber = 2
    Track3 TrackNumber = 3
)
```
Within this suggestion I also renamed the variables as `field.VersionFirst` to me doesn't sound related to track data.
What do you think?

field/track.go
```diff
@@ -0,0 +1,404 @@
...
+
+var (
+	expiryDateFormat   = "0601"
+	trackFirstRegex    = regexp.MustCompile(`^([A-Z]{1})([0-9]{1,19})\^([^\^]{2,26})\^([0-9]{4}|\^)([0-9]{3}|\^)([^\?]+)$`)
+	trackSecondRegex   = regexp.MustCompile(`^([0-9]{1,19})\=([0-9]{4}|\=)([0-9]{3}|\=)([^\?]+)$`)
+	trackThirdRegex    = regexp.MustCompile(`^([0-9]{2})([0-9]{1,19})\=([^\?]+)$`)
+	trackFirstPattern  = `^([A-Z]{1})([0-9]{1,19})\^([^\^]{2,26})\^([0-9]{4}|\^)([0-9]{3}|\^)([^\?]+)$`
+	trackSecondPattern = `^([0-9]{1,19})\=([0-9]{4}|\=)([0-9]{3}|\=)([^\?]+)$`
+	trackThirdPattern  = `^([0-9]{2})([0-9]{1,19})\=([^\?]+)$`
+)
```
louisheath, line 69: To avoid possible copy paste errors here we could consider using constants
```suggestion
const (
	expiryDateFormat   = "0601"
	
	track1Pattern = `^([A-Z]{1})([0-9]{1,19})\^([^\^]{2,26})\^([0-9]{4}|\^)([0-9]{3}|\^)([^\?]+)$`
	track2Pattern = `^([0-9]{1,19})\=([0-9]{4}|\=)([0-9]{3}|\=)([^\?]+)$`
	track3Pattern = `^([0-9]{2})([0-9]{1,19})\=([^\?]+)$`
)

var (
	track1Regex = regexp.MustCompile(track1Pattern)
	track2Regex = regexp.MustCompile(track2Pattern)
	track3Regex = regexp.MustCompile(track3Pattern)
)
```

field/track.go
```diff
@@ -0,0 +1,404 @@
...
+	return &Track{
+		spec:    spec,
+		Version: version,
+	}, nil
+}
+
+func NewTrack1Value(val []byte, version TrackVersion, isFixed bool) (*Track, error) {
+	if !version.Valid() {
+		return nil, errors.New("invalid track version")
+	}
```
louisheath, line 84: Do we need `version` to be a parameter if this function is specific to track 1?

field/track.go
```diff
@@ -0,0 +1,404 @@
...
+	return &Track{
+		spec:    spec,
+		Version: version,
+	}, nil
+}
+
+func NewTrack1Value(val []byte, version TrackVersion, isFixed bool) (*Track, error) {
```
louisheath, line 81: nit: I wasn't sure what `isFixed` meant at first
 ```suggestion
func NewTrack1Value(val []byte, version TrackVersion, fixedLength bool) (*Track, error) {
```

File: field/track.go [hunk 17]
```diff
@@ -0,0 +1,404 @@
...
+				return `=`
+			}
+			return raw
+		}
+	}
+
+	generator, _ := utils.NewGenerator(pattern, &utils.GeneratorArgs{
+		Flags:               syntax.Perl,
+		CaptureGroupHandler: handler,
+	})
```
louisheath, line 395: So I'm understanding `regen` to be a way of generating a random string using a regular expression,
and then using these handlers to inject strings we actually want at each index. This feels long winded.

Would something like `fmt.Sprintf("%s%s%s", a, b, c)` not suffice?

utils/regen.go
```diff
@@ -0,0 +1,244 @@
+package utils
+
+// https://github.com/zach-klippenstein/goregen // Apache-2.0 License
```
louisheath, line 3: I think this file warrants more documentation to help readers understand its purpose

louisheath (commented):
Thanks @mfdeveloper508, this PR is a massive help for us over at TM. Really excited to see this in master

(Apologies for the mass of separate comments, I haven't reviewed on Github in a while!)

field/track.go
```diff
@@ -0,0 +1,404 @@
...
+var cardTypes = map[string]string{
+	"Visa":             `^4[0-9]{12}(?:[0-9]{3})?$`,
+	"MasterCard":       `^5[1-5][0-9]{14}$`,
+	"American Express": `^3[47][0-9]{13}$`,
+	"Diners Club":      `^3(?:0[0-5]|[68][0-9])[0-9]{11}$`,
+	"Discover":         `^6(?:011|5[0-9]{2})[0-9]{12}$`,
+	"JCB":              `^(?:2131|1800|35\d{3})\d{11}$`,
+}
+
+func (f *Track) GetCardType() string {
```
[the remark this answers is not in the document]
mfdeveloper508: the issue is originated to adding utility functions for track
so I am trying to add some stuffs for track
I hope that the utility function will be use to check card type

krishishah:
As of https://github.com/moov-io/iso8583/pull/127/files, fields now should implement the `json.Unmarshaler` interface and `SetBytes()` now also sets bytes for the original data struct if one already exists. The changes to `field/composite.go` should give you a decent idea of what's involved there.

Could you update your PR to reflect this please? :pray:

mfdeveloper508:
> As of https://github.com/moov-io/iso8583/pull/127/files, fields now should implement the `json.Unmarshaler` interface and `SetBytes()` now also sets bytes for the original data struct if one already exists. The changes to `field/composite.go` should give you a decent idea of what's involved there.
> 
> Could you update your PR to reflect this please? 🙏

mfdeveloper508:
> > As of https://github.com/moov-io/iso8583/pull/127/files, fields now should implement the `json.Unmarshaler` interface and `SetBytes()` now also sets bytes for the original data struct if one already exists. The changes to `field/composite.go` should give you a decent idea of what's involved there.
> > Could you update your PR to reflect this please? 🙏

I see, will check again

field/track.go
```diff
@@ -0,0 +1,332 @@
...
+
+func (v TrackNumber) Valid() bool {
+	if v == Track1 || v == Track2 || v == Track3 {
+		return true
+	}
+	return false
+}
+
+const (
+	Track1 TrackNumber = 1
```
alovak, line 53: @mfdeveloper508 can we split Track into separate structs/fields: Track1, Track2 and Track3? It seems that it will simplify the code and make it easier to understand and maintain.

field/track.go
```diff
@@ -0,0 +1,332 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+	"time"
+)
+
+var _ Field = (*Track)(nil)
+
+type Track struct {
+	spec *Spec  `xml:"-" json:"-"`
```
alovak, line 14: why do we need `xml` tags?

wadearnold:
Hey, @alovak @krishishah @louisheath these are pretty small refactors. I agree with everything suggested. @mfdeveloper508 is grinding down some mission-critical issues for ISO20022. If we want to get this merged can some contribute some of these reactors so we can get this merged?

krishishah:
> Hey, @alovak @krishishah @louisheath these are pretty small refactors. I agree with everything suggested. @mfdeveloper508 is grinding down some mission-critical issues for ISO20022. If we want to get this merged can some contribute some of these reactors so we can get this merged?

I don't think the current implementation works for the typed API (which we use in our production setup) to pack, unpack etc. However, I understand that the PR is dragging on a bit. I'll take a step back from this PR and push a follow up to plug any gaps we see in the implementation.

Thanks!

field/track/track1.go
```diff
@@ -0,0 +1,202 @@
...
+	ExpirationDate       *time.Time  `json:"expiration_date,omitempty"`
+	ServiceCode          string      `json:"service_code,omitempty"`
+	DiscretionaryData    string      `json:"discretionary_data,omitempty"`
+}
+
+const (
+	expiryDateFormat = "0601"
+	track1Format     = `%s%s^%s^%s%s%s`
+	track2Format     = `%s=%s%s%s`
+	track3Format     = `%s%s=%s`
```
alovak, line 30:
```suggestion
```

field/track/track1.go
```diff
@@ -0,0 +1,202 @@
...
+	Name                 string      `json:"name,omitempty"`
+	ExpirationDate       *time.Time  `json:"expiration_date,omitempty"`
+	ServiceCode          string      `json:"service_code,omitempty"`
+	DiscretionaryData    string      `json:"discretionary_data,omitempty"`
+}
+
+const (
+	expiryDateFormat = "0601"
+	track1Format     = `%s%s^%s^%s%s%s`
+	track2Format     = `%s=%s%s%s`
```
alovak, line 29: let's move it to the related files

field/track/track1.go
```diff
@@ -0,0 +1,202 @@
...
+	track3Format     = `%s%s=%s`
+)
+
+var (
+	track1Regex = regexp.MustCompile(`^([A-Z]{1})([0-9]{1,19})\^([^\^]{2,26})\^([0-9]{4}|\^)([0-9]{3}|\^)([^\?]+)$`)
+	track2Regex = regexp.MustCompile(`^([0-9]{1,19})\=([0-9]{4}|\=)([0-9]{3}|\=)([^\?]+)$`)
```
alovak, line 35: let's move it to the related files

alovak:
@krishishah I've updated this PR and now it supports typed API as well. Also, tracks are now under the `field` package.

alovak:
I'm not sure about `NewTrackXValue` functions... I just removed them for a now as I don't see what is the value in them.

PR #127, opened 2021-09-22
Implement JSON Unmarshaler interface on Message and Field types

krishishah:
# New Behaviour

As of this PR, users may now unmarshal JSON back into an `*iso8583.Message` object using the untyped API as such:

```
input := `{"0":"0100","1":"500000000000000000000000000000000000000000000000","2":"4242424242424242","4":"100"}`

message := NewMessage(spec)
if err := json.Unmarshal([]byte(input), message); err != nil {
    // handle err
}
```

Similarly, the typed API may be used as such:

```
input := `{"0":"0100","1":"700000000000000000000000000000000000000000000000","2":"4242424242424242","3":{"1":"12","2":"34","3":"56"},"4":"100"}`

message := NewMessage(spec)
data := &TestISOData{}
if err := message.SetData(data); err != nil {
    // handle error
}
if err = json.Unmarshal([]byte(input), message); err != nil {
    // handle err
}

fmt.Printf("Field F2: %v", data.F2.Value) // This will print "4242424242424242"
```

This allows for some powerful behaviour to be built on top of the library e.g. Building JSON APIs which then convert messages into ISO8583 to then process for test purposes.

I've also improved some of the error handling and the implementation of `SetBytes()` to set the value of the data struct if one exists along the way.

### DIFF
--- a/field/binary.go
+++ b/field/binary.go
@@ -37,6 +37,9 @@ func (f *Binary) SetSpec(spec *Spec) {
 
 func (f *Binary) SetBytes(b []byte) error {
 	f.Value = b
+	if f.data != nil {
+		*(f.data) = *f
+	}
 	return nil
 }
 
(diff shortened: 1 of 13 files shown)


alovak (approved):
Great change! Thank you for your contribution!

PR #129, opened 2021-09-23
Composite fields json support

Lukepmcauley:
This PR add support for importing / exporting json specs that have composite fields. 
I've made the configuration for the builders public so users can specify any custom mappings they might have.

### DIFF
--- a/specs/builder.go
+++ b/specs/builder.go
@@ -115,8 +115,8 @@ type fieldDummy struct {
 	Enc         string                `json:"enc,omitempty" xml:"enc,omitempty"`
 	Prefix      string                `json:"prefix,omitempty" xml:"prefix,omitempty"`
 	Padding     *paddingDummy         `json:"padding,omitempty" xml:"padding,omitempty"`
-	Tag         *tagDummy             `json:"tag,omitempty"`
-	Subfields   map[string]*fieldDummy `json:"subfields,omitempty"`
+	Tag         *tagDummy             `json:"tag,omitempty" xml:"tag,omitempty"`
+	Subfields   map[string]*fieldDummy `json:"subfields,omitempty" xml:"subfields:omitempty"`
 }
 
 type paddingDummy struct {
--- a/specs/builder.go
+++ b/specs/builder.go
@@ -4,21 +4,21 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"reflect"
-	"sort"
-	"strconv"
-
 	"github.com/moov-io/iso8583"
 	"github.com/moov-io/iso8583/encoding"
 	"github.com/moov-io/iso8583/field"
 	"github.com/moov-io/iso8583/padding"
 	"github.com/moov-io/iso8583/prefix"
+	moovsort "github.com/moov-io/iso8583/sort"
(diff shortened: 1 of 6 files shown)


alovak:
@Lukepmcauley, currently, we have `Inspect` method for `Prefixer` and `Padder`. But we don't have it for `Encoder` and `StringSlice`. Our initial assumption was to use `Inspect` for `JSON` serialization, `describe`, etc. features. But as far as I see, now we use it only in `spec.Builder` and for things that don't have it, we use maps like `encodingsExtToInt`, `encodingsIntToExt`. As result, we have a mixed approach to a string representation.

What do you think about removing `Inspect` method at all, revering `StringSlice` changes, and just using maps/functions to get string representation inside `spec.Builder` for everything? It will streamline how we get a string representation and may reduce the number of changes required in the future.

One drawback of such change is that it's an API-breaking change (removing `Inspect()` method).

Thoughts?

@krishishah @vtolstov

Lukepmcauley:
@alovak For the sorts I can revert the changes and use the function name for serialisation. 
For the encoders and padders we'd still need some method of serialisation as type name is not enough information to recreate them and the builder doesn't have access to their internals. Any suggestions?

alovak:
Yeah. I agree. We can't get the meta-information for padder and prefixer. Then, let's keep it as is (with `Inspect()`) and move the rest into `spec.Builder` maps. How does it sound?

specs/builder.go
```diff
@@ -13,20 +13,21 @@ import (
 	"github.com/moov-io/iso8583/field"
 	"github.com/moov-io/iso8583/padding"
 	"github.com/moov-io/iso8583/prefix"
+	moovsort "github.com/moov-io/iso8583/sort"
 )
 
-type fieldConstructorFunc func(spec *field.Spec) field.Field
+type FieldConstructorFunc func(spec *field.Spec) field.Field
```
alovak, line 19: Why do you think we need it (and following) to be public?

Lukepmcauley: It allows library users to use their own custom fields, encodings, sort methods etc and then register them here before exporting their spec. Currently if you use a custom type the builder will complain.

alovak: That's fair.

vtolstov:
> @Lukepmcauley, currently, we have `Inspect` method for `Prefixer` and `Padder`. But we don't have it for `Encoder` and `StringSlice`. Our initial assumption was to use `Inspect` for `JSON` serialization, `describe`, etc. features. But as far as I see, now we use it only in `spec.Builder` and for things that don't have it, we use maps like `encodingsExtToInt`, `encodingsIntToExt`. As result, we have a mixed approach to get a string representation of the “thing”.
> 
> What do you think about removing `Inspect` method at all, revering `StringSlice` changes, and just using maps/functions to get string representation inside `spec.Builder` for everything? It will streamline how we get a string representation and may reduce the number of changes required in the future.
> 
> One drawback of such change is that it's an API-breaking change (removing `Inspect()` method).
> 
> Thoughts?
> 
> @krishishah @vtolstov

i don't use inspect method and i'm ok with breaking changes.

PR #133, opened 2021-10-03
Encoding / Decoding Bug Fixes

krishishah:
This PR consists of various encooding / decoding fixes to the library:
 1. all `field.JSONUnmarshal` methods now unquote values using the `json.Unmarshal` rather than using `strconv.Unquote`. This is because the `json.Unmarshal` function is more robust and handles a wider range of edge cases. 
 2. `field/Binary` `MarshalJSON` and `UnmarshalJSON` returns hex based strings and takes them in respectively. Previously, they handled bas64 representations of bytes making them pretty unusable in their string formats. This should be regarded to be a breaking change even though the API itself is unchanged.
 3. `field/Bitmap` no longer supports JSON encoding / decoding. The decision was made to do this because bitmaps add little value in JSON form and are used specifically for the parsing of ISO8583 messages with no business use case.
 4. `field/Composite` now lazily creates subfields only if they are being packed or unpacked rather than creating all subfields when defined in the spec regardless of whether they are to be populated with values or not. This change was necessary to ensure that only relevant fields are populated in their JSON format. Before, even unpopulated fields were populated with `""` or `null` values. This caused issues when unmarshalling in that JSON representation back into the `iso8583.Message`. 
 5. Replace `spec.CreateSubfields` with `field.CreateSubfield`. Due to the change in how `Composite` now creates fields, we needed to change the function signature of this function.

All of the above are required to deterministically Marshal a message to JSON, Unmarshal it back to a message, unpack it and pack it etc and get the same results regardless of the transformations


field/numeric_test.go
```diff
@@ -120,3 +120,32 @@ func TestNumericSetBytesSetsDataOntoDataStruct(t *testing.T) {
 
 	require.Equal(t, 9, data.Value)
 }
+
+func TestNumericJSONMarshal(t *testing.T) {
+	numeric := NewNumeric(&Spec{
```
alovak, line 125: I'm thinking if we can repalce it with:
```go
numeric := NewNumericValue(1)
```

It seems we don't do pack/unpack during marshaling and we may not need Spec here.

krishishah: nice find! I'll do the same for strings and binary too

field/binary.go
```diff
@@ -112,13 +113,23 @@ func (f *Binary) SetData(data interface{}) error {
 }
 
 func (f *Binary) MarshalJSON() ([]byte, error) {
-	return json.Marshal(f.Value)
+	str, err := f.String()
```
alovak, line 116: From what I see in `json` package:

```go
// Array and slice values encode as JSON arrays, except that
// []byte encodes as a base64-encoded string, and a nil slice
// encodes as the null JSON value.
```

and calling `f.String` here to get string representation of `Value` and then `encoding.ASCIIToHex.Encode` to get it back looks unnecessary.

Maybe we can just `json.Marshal(f.Value)` and then unmarshal it
```go
var []byte value
err := json.Unmarshal(b, &v)
return f.SetBytes(value)
```

Am I missing something?

alovak:
@krishishah thanks for the PR! It looks great!

PR #149, opened 2021-11-26
Return early when packing zero length fields

Dakinola892:
Currently, composite fields will error when they try and pack zero length fields, a situation which arises when not all subfields in a composite (w/out id tags) are present. 

The problem occurs when call .Pack() on individual fields, non populated fields would try and encode a zero length buffer and return an error - instead, now the field types will simply return an empty buffer and not error. 

This will allows the packing of composite fields with non-populated subfields that fall outside of the composite prefix length.

### DIFF
--- a/field/binary.go
+++ b/field/binary.go
@@ -65,6 +65,10 @@ func (f *Binary) Pack() ([]byte, error) {
 		return nil, fmt.Errorf("failed to encode content: %w", err)
 	}
 
+	if len(packed) == 0 {
+		return []byte{}, nil
+	}
+
 	packedLength, err := f.spec.Pref.EncodeLength(f.spec.Length, len(packed))
 	if err != nil {
 		return nil, fmt.Errorf("failed to encode length: %w", err)
--- a/field/composite_test.go
+++ b/field/composite_test.go
@@ -545,17 +545,26 @@ func TestCompositePackingWithTags(t *testing.T) {
 
 	t.Run("Pack correctly ignores excess subfields in excess of the length described by the prefix", func(t *testing.T) {
 		type ExcessSubfieldsTestData struct {
-			F1 *String
-			F2 *Numeric
-			F3 *String
-			F4 *String
-			F5 *String
+			F1  *String
+			F2  *Numeric
+			F3  *String
+			F4  *String
+			F5  *String
+			F6  *String
+			F7  *String
+			F8  *String
+			F9  *Numeric
+			F10 *String
+			F11 *String
+			F12 *String
+			F13 *String
+			F14 *String
 		}
 
 		excessSubfieldsSpec := &Spec{
-			Length:      5,
+			Length:      26,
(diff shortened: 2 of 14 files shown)


wadearnold (approved):
Thanks for the contributions!

field/binary.go
```diff
@@ -65,6 +65,10 @@ func (f *Binary) Pack() ([]byte, error) {
 		return nil, fmt.Errorf("failed to encode content: %w", err)
 	}
 
+	if len(packed) == 0 {
```
alovak, line 68: I think that makes sense! Not sure why we didn't add it before :)
Also, it would be great to see a test for checks in this and other files. Just test that for a zero value it returns an empty slice.

field/composite_test.go
```diff
@@ -543,6 +543,80 @@ func TestCompositePackingWithTags(t *testing.T) {
...
+			},
+		}
+
+		data := &ExcessSubfieldsTestData{}
+
+		composite := NewComposite(excessSubfieldsSpec)
+		err := composite.SetData(data)
+		require.NoError(t, err)
+
+		// Subfield 4 & 5 falls outside of the bounds of the 3 byte limit imposed
```
alovak, line 602: Also, I'm not sure I understand this bit of the PR. How does the early return help with fields that fall outside of the bounds? Also, where 3 byte limit prefix is? Is it because of `ebcdic1047`? Sorry for the dumb question.

Dakinola892: Hi, this probably wasn't explained the best by me - the motivation behind this PR was to allow  length-prefixed composites that with non-tagged subfields to be parsed only up to the specified length, so that non-populated subfields in the composite could be ignored without erroring/panicking.

The problem with the current implementation is when trying to parse a composite without tagged subfields, Composite loops through all the subfield IDs in the spec regardless and errors when it tried to unpack a zero-length data buffer for a subfield not included in the data.  

A possible fix would have been to just add check to Composite to stop parsing when we'd read >= the length of the data, but this caused an issue where errors weren't being accurately reported for invalid length data that was too long (i.e the spec only allows 4 bytes but 7 bytes were provided, it would stop reading after >= 4 bytes, so it wouldn't accurately highlight the issue with the combined length of the subfields). 

So now I attempted to get the desired result by focusing on not panicking when attempting to unpack zero-length data fields.

Dakinola892: In the original example, I was referring to the '03' prefix when mentioning the 3 byte limit (i.e. [03 | 102] -> [length prefix | data], but I've updated it the most recent commit to hopefully make it clearer

field/binary.go
Dakinola892: No problem - added a small test now

adamdecaf (approved):

adamdecaf:
I'm looking for an approval from @alovak before we merge.

alovak (approved):

PR #150, opened 2021-11-26
Add Message .Clone()

Dakinola892:
This is a basic implementation of a method used to clone a Message - it creates a new basic message, sets the spec from the given message and populates the new message with the packed message bytes from the given message. 

This should prove useful for constructing response messages which share the majority of fields with their requests. The necessary fields of the cloned requests can be altered to form a response.

### DIFF
--- a/message.go
+++ b/message.go
@@ -344,3 +344,20 @@ func (m *Message) setUnpackableDataField(id int) error {
 func (m *Message) dataFieldValue(id int) reflect.Value {
 	return m.dataValue.FieldByName(fmt.Sprintf("F%d", id))
 }
+
+func (m *Message) Clone() (*Message, error) {
+	newMessage := NewMessage(m.GetSpec())
+	newMessage.SetData(reflect.New(m.dataValue.Elem().Type()))
+
+	bytes, err := m.Pack()
+	if err != nil {
+		return nil, err
+	}
+
+	err = newMessage.Unpack(bytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return newMessage, nil
+}
--- a/message.go
+++ b/message.go
@@ -251,7 +251,9 @@ func (m *Message) MarshalJSON() ([]byte, error) {
 
 func (m *Message) UnmarshalJSON(b []byte) error {
 	var data map[string]json.RawMessage
-	json.Unmarshal(b, &data)
+	if err := json.Unmarshal(b, &data); err != nil {
+		return err
+	}
 
 	for id, rawMsg := range data {
 		i, err := strconv.Atoi(id)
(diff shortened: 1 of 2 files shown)


wadearnold (approved):
This is great.

wadearnold (changes requested):
I get a panic when running the test cases

alovak:
@Dakinola892 thanks for the PR!

I think that this is a great start for the `Clone` method. I see `Clone` method has to clone everything inside the message. While spec can be re-used (we do not modify it), the message fields and other internal things should be copied in a similar way that is implemented here: https://github.com/barkimedes/go-deepcopy

There are some areas we have to pay attention to:
* Data - we link message fields with struct fields. It means that if we just copy pointers, changing data in cloned message will change the data in the original message.
* message fields that store slices - we need to re-create slice and copy data from one slice to a new one
* we have to clone bitmap
* properly clone composite fields
* ...

what do you think about it all?

alovak:
@Dakinola892 I was thinking about something like https://github.com/barkimedes/go-deepcopy but your approach is simpler and it works! Thanks for the contributions!

Dakinola892:
Unfortunately, i think I missed something when initially running the tests because I've been looking into it further and my implementation attempt is a bit to naive. Handling the deep copies of the all the fields is a challenge, especially dealing with composites since extracting the custom structs used to populate them isn't the simplest.

I wouldn't be against using go-deepcopy as an easier avenue if there is no pushback with relying on a 3rd-party library

Dakinola892:
@alovak I experimented with go-deepcopy, and it doesn't seem to be able to handle composite fields (attempting to deepcopy.Anything() times out after reaching max depth of pointer loop), so I tried to fix my current implementation. 

These new changes creates allow .Clone() to return a new message that has an identical spec, bitmap and field values, however it is not completely identical to the original method. The cloned messages' fields do not have a data value or spec in the Field itself. Currently .Clone() will return identical messages for messages initially populated via .Unpack() or .UnmarshalJSON() (since the fields in the original also don't have spec & data fields populated), but not for those populated via SetData() with a populated Data struct.

I'm curious on your opinion if this matters or not, as the usecase for a .Clone() function is mostly to have an identical, mutable object with identical accessible field values, and I believe these changes accomplish that, so it may not be necessary to further pursue a perfect copy of messages.

vtolstov:
i'm not sure... but why to encode to buffer prev message and decode bytes from buffer to new message ?
this creates new copy of message..

alovak:
@vtolstov not sure I understand your question. Do you agree that it is enough to pack/unpack to clone a message? Or you have other ideas?

vtolstov:
> @vtolstov not sure I understand your question. Do you agree that it is enough to pack/unpack to clone a message? Or you have other ideas?

yes, pack/unpack will create new message without pain via reflect and other stuff. So i think that internally pack/unpack will be good

Dakinola892:
That’s the idea – the reflect package is only used to set a blank struct using .SetData() so the values can be access via the typed API, but the Fields won’t have their own spec or data values, only Value (or equivalent for Composites). From: Vasiliy TolstovSent: 02 December 2021 22:34To: moov-io/iso8583Cc: Daniel Akinola; MentionSubject: Re: [moov-io/iso8583] Add Message .Clone() (PR ***@***.*** not sure I understand your question. Do you agree that it is enough to pack/unpack to clone a message? Or you have other ideas?yes, pack/unpack will create new message without pain via reflect and other stuff. So i think that internally pack/unpack will be good—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android.

wadearnold (approved):

adamdecaf (approved):

adamdecaf:
I'm looking for an approval from @alovak before we merge.

alovak (approved):
Let's try this out!

PR #151, opened 2021-12-01
Fix error wrapping

javorkolev123:
The PR fixes an issue with the error wrapping in different places across the repository. 
%w should be used as otherwise errors can't be properly unwrapped when using functions
such as error.As().


javorkolev123:
Merge conflicts require resolving.

PR #152, opened 2021-12-02
Replace %v with %w in error messages

javorkolev123:
A small fix that would allow us to unwrap go errors properly by using %w instead %v when generating wrapped errors.


wadearnold (approved):
Thank you! Thats on me!

alovak (approved):
Beautiful! Thanks for the contribution!

PR #157, opened 2022-01-21
implement GetData

alovak:
##What

With this PR we make it possible to get message fields data into Go struct by adding `message.GetData(data)`. It works in the same way as `SetData` + `Data()`.

##Why

The current implementation of how we get typed data does not allow us to get message field values into struct fields without calling `Unpack` before.

Currently, `SetData` method links struct fields with the message field values and sets struct field values only when `Pack` and `Unpack` methods are called. You have to SetData first, call `Unpack` second, and only after that your structure will receive values.

Because getting data in linked with unpacking we can't get data after message was unpacked.

example of current behavior:

```
// we define type for data we want to work with
type ISO87Data struct {
	F0 *field.String
	F2 *field.String
	F4 *field.String
}

message := NewMessage(spec)

// we link message fields with struct fields
message.SetData(&ISO87Data{})

// we unpack message
rawMsg := []byte("01007000000000000000164242424242424242123456000000000100")
err := message.Unpack([]byte(rawMsg))

require.NoError(t, err)

// now we can get data
data := message.Data().(*ISO87Data)

require.Equal(t, "0100", data.F0.Value)
require.Equal(t, "4242424242424242", data.F2.Value)
require.Equal(t, "12", data.F3.F1.Value)
require.Equal(t, "34", data.F3.F2.Value)
require.Equal(t, "56", data.F3.F3.Value)
require.Equal(t, "100", data.F4.Value)
```

next actions:
* Remove `message.Data()`
* Rework `SetData` to set field values only when we call it. Do not use it during `Pack` and `Unpack`


adamdecaf:
This makes sense and does simplify the API, but since it breaks the API I'd like to get some input from community folks who use this library as well. 

The SetData -> Unpack() thing confused me as well in the past.

alovak:
@adamdecaf I agree. Currently, we do not break API. `SetData` and `Data` will continue to work as they worked before, but I want to change them in the next releases.

wadearnold:
Can we add the deprecated comment to `Data()` in this pull request. That would allow linters to catch the change before we break the api. 

```go
// Deprecated
```

wadearnold (approved):

field/composite.go
```diff
@@ -89,6 +90,49 @@ func (f *Composite) SetSpec(spec *Spec) {
 	f.orderedSpecFieldTags = orderedKeys(spec.Subfields, spec.Tag.Sort)
 }
 
+func (f *Composite) GetData(data interface{}) error {
+	rv := reflect.ValueOf(data)
+	if rv.Kind() != reflect.Ptr || rv.IsNil() {
+		return errors.New("data is not a pointer or nil")
+	}
```
louisheath, line 97: Should we return nil if `rv.IsNil()`? Will this panic otherwise? Could be worth a unit test

field/composite.go
```diff
@@ -89,6 +90,49 @@ func (f *Composite) SetSpec(spec *Spec) {
...
+	if dataStruct.Kind() != reflect.Struct {
+		return errors.New("data is not a struct")
+	}
+
+	// iterate over struct fields
+	for i := 0; i < dataStruct.NumField(); i++ {
+		dataFieldName := dataStruct.Type().Field(i).Name
+
+		// skip struct field if its name starts not from F
+		if dataFieldName[0:1] != "F" {
```
louisheath, line 111: Can we nil dereference panic here?

message.go
```diff
@@ -65,6 +67,56 @@ func (m *Message) SetData(data interface{}) error {
 	return nil
 }
 
+func (m *Message) GetData(data interface{}) error {
+	rv := reflect.ValueOf(data)
+	if rv.Kind() != reflect.Ptr || rv.IsNil() {
+		return errors.New("data is not a pointer or nil")
+	}
```
louisheath, line 74: Same comment as in `composite.go` around the nil check

message.go
```diff
@@ -65,6 +67,56 @@ func (m *Message) SetData(data interface{}) error {
...
+
+	// iterate over struct fields
+	for i := 0; i < dataStruct.NumField(); i++ {
+		dataFieldName := dataStruct.Type().Field(i).Name
+
+		fmt.Printf("dataFieldName = %+v\n", dataFieldName)
+
+		// skip struct field if its name starts not from F
+		if dataFieldName[0:1] != "F" {
```
louisheath, line 90: Likewise for nil dereference here

cheukwing (commented):
Thanks! I think this a good change, but I agree with Louis that `GetData` (as well as `SetData`) feel a bit off naming-wise. Again, not bothered if you decide to keep it as it is though.

alovak:
@wadearnold I completely agree with removing Get for getters. Somehow I missed the go recommendation and we have opposite naming in the package: `message.Field(1, "value")` and `message.GetField(1)`. I hope we can fix it before we release v1.

For `GetData` and `SetData` - it's trickier as they are not just a field getter or a field setter.
We are setting values for the message fields taking them from the corresponding field of the input struct for `SetData` and doing the opposite for `GetData`. Should it be something like `Populate` but here is unclear the direction of the action...

>    There are only two hard things in Computer Science: cache invalidation and naming things.
    -- Phil Karlton

any other ideas for `GetData`, `SetData` alternative names?

alovak:
@cheukwing @louisheath what do you guys think about `Values/SetValues` or `FieldValues/SetFieldValues`?

cheukwing:
I think I would assume that something called `FieldValues` would return the values, rather than having to give it a struct which it would then fill via reference.
A verb like `Populate`, `Unmarshal`, `Fill` etc might be nicer, but again, not a big deal if we just go with `FieldValues` (`Data/Values` might be a bit too vague?)

atonks2:
I agree with @cheukwing. This function isn't getting/returning data, it's unmarshaling it. I believe `Unmarshal` would be the most clear/intuitive. @adamdecaf 's suggestion also makes a lot of sense, but I wonder how it'd handle subfields. Simplest way forward seems to be `Unmarshal`.

field/composite.go
alovak: @louisheath not sure I understand your concern. We expect that data must be a pointer to the struct, otherwise we can't set values. The trick with `iso8583.Unmarshal` (old `GetData`) is that if field in the struct is not set, it will instantiate an empty value:

```
if dataField.IsNil() {
	dataField.Set(reflect.New(dataField.Type().Elem()))
}
```

Am I missing something here?

alovak:
I've updated the PR. I addressed the feedback and now we have:

```
data := &authData{}

iso8583.Unmarshal(message, data)
```

Is `Unmarshal` good? Should it be `UnmarshalValues`? :)

decode.go
```diff
@@ -0,0 +1,60 @@
...
+
+		// skip struct field if its name starts not from F
+		if len(dataFieldName) == 0 || dataFieldName[0:1] != "F" {
+			continue
+		}
+
+		indexStr := dataFieldName[1:]
+		fieldIndex, err := strconv.Atoi(indexStr)
+		if err != nil {
+			return errors.Wrap(err, "converting field intex into int")
```
cheukwing, line 39: Still need to replace this with `fmt.Errorf`.

We can import the stdlib `errors` instead of `pkg/errors` and still have `errors.New` functionality, without adding a new dependency.

cheukwing (approved):
Thanks! I think just `Unmarshal` is clear enough for naming.

Let's remember to remove that new dependency though.

decode.go
alovak: Oh yeah. Fixed it! Thanks!

atonks2 (approved):

field/composite.go
louisheath: I was naively drawing parallels from similar implementations which do
```
	if v == nil {
		return nil
	}
```
But it appears the requirements are different here

louisheath (approved):
Happy with @cheukwing 's review here.
Thanks all for the discussion and apologies for dragging out this PR! I think it was worthwhile.

Issue #159, opened 2022-01-28
Panic on deserialising a short message


peterebden:
`Message.Unpack()` panics if given an input shorter than 4 bytes:

```
	t.Run("Test unpacking an empty input", func(t *testing.T) {
		message := NewMessage(spec)
		message.Unpack(nil)
	})
```

```
--- FAIL: TestMessage (0.00s)
    --- FAIL: TestMessage/Test_unpacking_an_empty_input (0.00s)
panic: runtime error: slice bounds out of range [:4] with capacity 3 [recovered]
	panic: runtime error: slice bounds out of range [:4] with capacity 3

goroutine 25 [running]:
testing.tRunner.func1.2({0x6130a0, 0xc0000ae438})
	testing/testing.go:1360 +0x24e
testing.tRunner.func1()
	testing/testing.go:1363 +0x39f
panic({0x6130a0, 0xc0000ae438})
	runtime/panic.go:838 +0x207
github.com/moov-io/iso8583/encoding.asciiEncoder.Decode({}, {0xc0001310f8?, 0x0?, 0x7ff419b0a5e0?}, 0x8?)
	github.com/moov-io/iso8583/encoding/ascii.go:23 +0x145
github.com/moov-io/iso8583/field.(*String).Unpack(0xc0000e5420, {0xc0001310f8, 0x3, 0x3})
	github.com/moov-io/iso8583/field/string.go:84 +0xf3
github.com/moov-io/iso8583.(*Message).Unpack(0xc000098fc0, {0xc0001310f8, 0x3, 0x3})
	github.com/moov-io/iso8583/message.go:186 +0x256
github.com/moov-io/iso8583.TestMessage.func4(0x0?)
	github.com/moov-io/iso8583/message_test.go:189 +0x4c
testing.tRunner(0xc000164ea0, 0xc000097740)
	testing/testing.go:1410 +0x102
created by testing.(*T).Run
	testing/testing.go:1457 +0x35f
exit status 2
FAIL	github.com/moov-io/iso8583	0.010s
```

PR #161, opened 2022-02-09
Fix panic on unpack nil

alovak:
In this PR we check input slice length before working with it to avoid panic errors: slice bounds out of range.

solves #159

### DIFF
--- a/encoding/ascii.go
+++ b/encoding/ascii.go
@@ -1,6 +1,8 @@
 package encoding
 
-import "fmt"
+import (
+	"fmt"
+)
 
 var ASCII Encoder = &asciiEncoder{}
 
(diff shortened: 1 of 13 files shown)


encoding/ascii.go
```diff
@@ -20,6 +22,9 @@ func (e asciiEncoder) Encode(data []byte) ([]byte, error) {
 
 func (e asciiEncoder) Decode(data []byte, length int) ([]byte, int, error) {
 	// read only 'length' bytes (1 byte - 1 ASCII character)
+	if len(data) < length {
+		return nil, 0, fmt.Errorf("not enought data to decode. expected len %d, got %d", length, len(data))
```
peterebden, line 26: bit of a spelling nit but `enought` -> `enough` :)

(and below)

peterebden (commented):
Thank you!

encoding/ascii.go
alovak: thanks for noticing it!

PR #168, opened 2022-02-28
use index tag first, field name second

alovak:
When we marshal/unmarshal field values, we match struct fields by name first and by `index` tag second. When we check the field name we use [regexp](https://github.com/moov-io/iso8583/blob/master/message.go#L401): `^F\d+$` (field name starts with `F` and has digits after it). Because composite fields can have tag names that are not digits, we only check that field starts with `F` with [the regexp](https://github.com/moov-io/iso8583/blob/master/message.go#L401): `^F.+$`.

So, if you marshal/unmarshal composite field that starts with `F` and also has index, then the index tag will be ignored. Imagine you have tag `01` and a struct:

```go
type OriginalElements struct {
        ForwadingInsitutionCode *field.String `index:"01"`
}
```

Guess what? You will spend time trying to understand why tag `01` is not set when you marhsal/unmarshal data. And it's because the composite field will take `orwadingInsitutionCode` as an index of the field and will try to find a spec for such field.

To avoid such cases, we are changing the field name matching priorities by checking the `index` tag first and the field name second.

**Technically, it can be a breaking change if you use `index` tag on the struct for something else.**

### DIFF
--- a/field/composite.go
+++ b/field/composite.go
@@ -476,12 +476,12 @@ var fieldNameTagRe = regexp.MustCompile(`^F.+$`)
 func getFieldIndexOrTag(field reflect.StructField) (string, error) {
 	dataFieldName := field.Name
 
-	if len(dataFieldName) > 0 && fieldNameTagRe.MatchString(dataFieldName) {
-		return dataFieldName[1:], nil
-	}
-
 	if fieldIndex := field.Tag.Get("index"); fieldIndex != "" {
 		return fieldIndex, nil
+	}
+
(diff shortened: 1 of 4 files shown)


adamdecaf (approved):

adamdecaf:
While this is a breaking change this PR resolves a pretty annoying bug that would take a while for someone to investigate, so it makes sense to fix. Also the original change is ~2 weeks old in this repository so few people have likely adopted it.

PR #169, opened 2022-03-01
filter PAN when describe message

alovak:


describe.go
```diff
@@ -13,6 +14,13 @@ import (
...
 var defaultSpecName = "ISO 8583"
 
+var panFilter = regexp.MustCompile(`^\d(\d*)\d{4}$`)
+
+const (
+	bitmapIndex = 1
+	panIndex    = 2
+)
+
 func Describe(message *Message, w io.Writer) error {
```
adamdecaf, line 24: FYI human-readable output typically isn't protected from breaking changes, so we're free to alter this as needed. Hiding PII is very important. 

We offer multiple machine readable formats which we promise to not break.

wadearnold:
Does this seem like it could get merged to filter PII data? What do you think @adamdecaf @alovak

describe_test.go
```diff
@@ -48,7 +48,7 @@ MTI...........................: 0100
 Bitmap........................: 400000000000000000000000000000000000000000000000
 Bitmap bits...................: 01000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
 F000 Message Type Indicator...: 0100
-F002 Primary Account Number...: 4242424242424242
+F002 Primary Account Number...: 4****4242
```
adamdecaf, line 51: I think the output should match the original length. PANs are pretty known lengths, so we aren't revealing much by showing the actual length.

alovak:
Implemented in https://github.com/moov-io/iso8583/pull/203

PR #180, opened 2022-06-27
Return error with invalid length

mfdeveloper508:
removed packed check, update composite field


field/composite.go
```diff
@@ -342,6 +338,20 @@ func (f *Composite) UnmarshalJSON(b []byte) error {
 }
 
 func (f *Composite) pack() ([]byte, error) {
+
+	// getting max length of data
+	encodedLength, err := f.spec.Pref.EncodeLength(f.spec.Length, f.spec.Length)
```
adamdecaf, line 343: Is this right?

mfdeveloper508: yes
added comment for this

the code is to get real data size without length prefix
length prefix of variable field is formatted string with same size, L/LL/LLL
so we can get real size of length prefix from spec length (instead of len(packed))

field/composite_test.go
```diff
@@ -542,6 +539,49 @@ func TestCompositePackingWithTags(t *testing.T) {
 		require.EqualError(t, err, "failed to encode length: field length: 12 should be fixed: 6")
 	})
 
+	t.Run("Pack returns error when encoded data length is larger than specified variable max length", func(t *testing.T) {
```
alovak, line 542: there was a test for fixed length, I added one for variable length.

field/composite_test.go
```diff
@@ -595,142 +635,6 @@ func TestCompositePackingWithTags(t *testing.T) {
...
-		// Therefore, it won't be included in the packed bytes.
-
-		packed := []byte("01110000100012")
-
-		read, err := composite.Unpack(packed)
-
-		require.NoError(t, err)
-		require.Equal(t, 14, read)
-
-		packedBytes, err := composite.Pack()
```
alovak, line 729: Now we just pack it back. Unset fields will not be included in the packed value. It's just expected behavior.

So, test is removed as we don't test here anything. Or I'm missing something.

alovak:
@adamdecaf please, take a look.

@mfdeveloper508 did what we had to do: just removed the check. It packs and unpacks according to expectations:
- If there is no data or more data for the fixed-length field - it will fail
- If there is more data than max variable length specified for the field - it will fail

I may miss some use-cases @Dakinola892  @cheukwing please review it as well.

I removed some unnecessary code and added some tests.

Dakinola892 (approved):
Happy with these new changes

PR #182, opened 2022-07-01
auto bitmaps size

mfdeveloper508:
- remove maxBitmaps
- use auto size according to specification
- added over range check
- updated test codes


adamdecaf:
This LGTM but I'd like @alovak to review it when he's back.

field/bitmap.go
```diff
@@ -170,15 +181,24 @@ func (f *Bitmap) bitmapsCount() int {
 	return count
 }
 
+func (f *Bitmap) getMapSize() int {
+	if f.mapSize == 0 {
+		f.mapSize = defaultBitmaps
```
Dakinola892, line 186: I think also a `f.Reset()` after this as well just in case - i realize that in practice that it will probably be used in conjuction with `setMapSize` which will call `f.Reset` but to me it feels a bit 'risky' to have a bitmap say it has size 1 but with no actually initialized values

mfdeveloper508: Yes, 
the code has problem when using Bitmap directly

wadearnold:
Hey, @alovak can you take a look at this?

alovak:
Sorry for the delay. I hope to review and merge it this/next week @mfdeveloper508 @wadearnold

wadearnold:
Can we get this merged @alovak ?

alovak:
The main issue with this implementation is that we should manually set the bitmap size, while in some cases we may not know it. The current implementation automatically reads the next bitmap if the first bit of the current bitmap is set and the solution suggested in the PR breaks this feature.

We should not do/have things like `bitmap.SetMapSize(2)` [here](https://github.com/moov-io/iso8583/pull/182/files#diff-355963d4026cc1676601433fe8aedcd4037905ff315d4d3763dc3508f1de962eR35) as we don't actually know the length of the bitmap. The bitmap should automatically read two bitmaps.

The alternative version is here: https://github.com/moov-io/iso8583/pull/211

PR #185, opened 2022-07-20
Wrap errors from external libraries to prevent leaking sensitive information

### LINKED ISSUE #103: Ensure sensitive information is not exposed through error messages
I've noticed with debugging some code that this library sometimes exposes sensitive information via error messages. An example of this is if `strconv.Atoi(...)` returns an error as part of `Numeric.Unpack(...)`. 

The library wraps around the error and adds additional context. However, `strconv.Atoi(...)` returns the value of the string that it could not convert to an int. Sensitive information could live inside of this field and may be accidentally exposed in logs.

We should ensure that errors from third party dependencies are handled such that this information would not be propagated. Same applies to errors generated directly by the library.

cheukwing:
Resolves #103 

Introduce a `SafeError` type (credits to @alovak for the idea!) which is used to wrap around external errors, preventing the returned error message from displaying sensitive information, while still allowing errors to be matched.
Use this new error type to wrap external errors in the `field` and `encoding` packages, as these operate on the potentially sensitive data.

I considered also wrapping external errors in the `prefix` and `network` packages, but since these only operate on the length part of the data, exposing their details _should_ be okay (as long as the message is correctly formatted).

### DIFF
--- a/encoding/ascii.go
+++ b/encoding/ascii.go
@@ -2,6 +2,8 @@ package encoding
 
 import (
 	"fmt"
+
+	"github.com/moov-io/iso8583/utils"
 )
 
 var ASCII Encoder = &asciiEncoder{}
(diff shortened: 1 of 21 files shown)


adamdecaf (approved):
I'm a fan of this, nice!

alovak (approved):
Great job! Thanks a lot!!!

PR #190, opened 2022-07-29
Add EBCDIC 1047 encoder & prefixer

Dakinola892:
EBCDIC has [many code pages](https://cs.opensource.google/go/x/text/+/refs/tags/v0.3.7:encoding/charmap/tables.go) with slightly different encodings for characters. 

[Code Page 1047](http://www.kostis.net/charsets/ebc1047.htm) is a commonly used EBCDIC spec with encodings that differ from the EBCDIC encoder already provided, so an additional encoder to handle encoding to this spec might be useful addition

### DIFF
--- a/encoding/ebcdic1047.go
+++ b/encoding/ebcdic1047.go
@@ -1,0 +1,35 @@
+package encoding
+
+import (
+	"github.com/moov-io/iso8583/utils"
+	xencoding "golang.org/x/text/encoding"
+	"golang.org/x/text/encoding/charmap"
+)
+
+// EBCDIC1047 is an encoder for EBCDIC characters using IBM Code Page 1047.
+var EBCDIC1047 Encoder = &ebcdic1047Encoder{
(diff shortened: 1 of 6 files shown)


encoding/ebcdic1047.go
```diff
@@ -0,0 +1,35 @@
...
+func (e ebcdic1047Encoder) Encode(data []byte) ([]byte, error) {
+	bytes, err := e.encoder.Bytes(data)
+	if err != nil {
+		return nil, utils.NewSafeError(err, "failed to encode EBCDIC")
+	}
+	return bytes, nil
+}
+
+func (e ebcdic1047Encoder) Decode(data []byte, length int) ([]byte, int, error) {
+	data = data[:length]
```
alovak, line 29: let's add a data length check here (you can quickly copy it from other encoders)

Dakinola892: good idea - added a length check now

alovak (approved):
Thank you!!!

PR #193, opened 2022-09-15
Fix variable length HEX prefix

svz-ya:
I believe that variable length HEX prefix is encoded/decoded incorrectly. For example, if we have a data block with length 24 bytes and encode `prefix.Hex.LLL` prefix it gives:
```
+ 00000000  33 32 33 34                                       |3234|
```
i.e. ASCII-HEX representation of decimal number `24` without even a padding to 3 digits.
However it is expected to have
```
- 00000000  30 30 30 30 31 38                                 |000018|
```
ASCII representation of `0x000018`. LLL means 3 bytes for hex number, which is 6 hex nibbles.

### DIFF
--- a/prefix/hex_test.go
+++ b/prefix/hex_test.go
@@ -3,10 +3,11 @@ package prefix
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func TestHex(t *testing.T) {
+func TestFixedHex(t *testing.T) {
 	pref := hexFixedPrefixer{}
 
 	dataLen, read, err := pref.DecodeLength(16, []byte("whatever"))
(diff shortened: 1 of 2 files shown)


prefix/hex_test.go
```diff
@@ -15,3 +16,155 @@ func TestHex(t *testing.T) {
...
+		{
+			name: "success",
+			fields: fields{
+				Digits: 3,
+			},
+			args: args{
+				maxLen: 32,
+				data:   []byte("000018whateverwhateverwhatever"),
+			},
+			wantDataLen: 0x18,
```
alovak, line 112: I would use `24` here as it is what we want to see

alovak:
@svz-ya I think that you are right. Unfortunately there are no tests for the hex prefixes. We should be careful with not breaking what may for others (even if we now see it's not how it should work 🙈)

prefix/hex_test.go
svz-ya: I intentionally use here `0x18` so a reader can match prefix `000018` on line 110 with expected length without making math calculation in mind.

svz-ya:
@alovak 

> We should be careful with not breaking what may for others (even if we now see it's not how it should work see_no_evil)

I agree. So what are your suggestions?

alovak:
@svz-ya please, rebase, and we will merge it :)

alovak:
@svz-ya I've rebased and merged it myself here: https://github.com/moov-io/iso8583/pull/198

thank you for your contribution!

alovak:
it was released https://github.com/moov-io/iso8583/releases/tag/v0.12.2 @svz-ya

PR #194, opened 2022-09-21
fix go 1.19 formatting

alovak:
What:
* `go fmt` two files as go 1.19 has changed how it formats comments and we have a linter that tells us we have to re-format some files


adamdecaf (approved):
Outside of the obvious ioutil deprecations.

Issue #195, opened 2022-09-22
Set empty value (zero value) for missing fields when unmashaling message into data struct


alovak:
# The problem

Let's imagine we have a following type:

```go
type NetworkManagementRequest struct {
	MTI                  *field.String `index:"0"`
	TransmissionDateTime *field.String `index:"7"`
	STAN                 *field.String `index:"11"`
	InformationCode      *field.String `index:"70"`
}
```

When we unmarshal message fields into the data struct before we access the field value, we should check if field is not nil:

```go
data := &NetworkManagementRequest{}
err := message.Unmarshal(data)
// handle error

// it's not always safe to access field value like this:
fmt.Printf("MTI: %s\n", data.MTI.Value)

// if field is not set in the message, then the corresponding field in the data struct will be nil
fmt.Printf("InformationCode": %s\n", data.InformationCode.Value) // => will panic

// so, to avoid this
if data.InformationCode != nil {
	// do something with the value
	data.InformationCode.Value
}
```

# Proposal for the solution
We can add  a value for the field tag e.g., `addempty` (it's like the opposite of `omitempty` for `json` tag), which will set an empty/zero value to the field when we message.Unmarshal(data) and the field was not in the message.

Example:

```go
type NetworkManagementRequest struct {
	MTI                  *field.String `index:"0"`
	TransmissionDateTime *field.String `index:"7"`
	STAN                 *field.String `index:"11"`
	InformationCode      *field.String `index:"70,addempty"`
}
```

Please, feel free to propose a better name instead of `addempty`.

P.S. having a tag with the name `index` looks weird, I think it's time to introduce the `iso8583` tag and keep `index` as an alias for backward compatibility.

adamdecaf:
`addempty` is a handy name for always populating `*field.T` values, so that makes sense to me. 

Is there a difference between adding this tag and guarding the accessor better? I understand this proposal would remove the possibility of nil panics on `.Value` access requests. Maybe both changes are needed?

We could guard the accessor methods better to avoid panics. See https://github.com/moov-io/iso8583/pull/196

alovak:
I think guarding accessor is something we definitely should add! Because in many places, we access `.Value` (as it's more convenient because `String()` returns two values), we should add `addempty` too.

alovak:
What if we add `Value()` method for nil receiver will return zero value? Then we don't need  to add `addempty`... we will just replace `posData.Field.Value` with `posData.Field.Value()`

alovak:
[...]

PR #196, opened 2022-09-22
field: nil check Numeric, String, Composite, etc accessors

adamdecaf:
Issue: https://github.com/moov-io/iso8583/issues/195


alovak (approved):

alovak:
This is a really good change to make code simpler for the callers

adamdecaf:
@alovak Cool. I can add this to every field type.

alovak:
thanks @adamdecaf! this PR looks 💪

the only thing that (kind of) worries me is that **Get**Value. On `message` we have `GetString()`, `GetBytes()`. But on `field` we have `String()` and `Bytes()` and now we add `GetValue()`.

We can introduce a breaking change by adding `Value()` and `SetValue()` and renaming `Value` into `value`. It will make code more consistent and safe. In my personal opinion, such change is just find/replace so it should not be a PITA.

thoughts?

adamdecaf:
Yea I agree that `GetValue()` isn't idea and I'm okay with that breaking change. It's minor enough and increases runtime safety.

adamdecaf:
@alovak feel free to merge this whenever you want. I know there are other 8583 changes in flight.

PR #203, opened 2022-11-21
new filters for specified fields

mfdeveloper508:


alovak:
@mfdeveloper508 thanks a lot! This looks great!

field_filter.go
```diff
@@ -0,0 +1,112 @@
...
+var Track3Filter = func(in string, data field.Field) string {
+	track := field.Track3{}
+	if err := newTrackData(data, &track); err != nil {
+		return in
+	}
+	track.PrimaryAccountNumber = PANFilter(track.PrimaryAccountNumber, nil)
+	return getTrackDataString(in, &track)
+}
+
+func newTrackData(data, track field.Field) error {
```
alovak, line 95: am I right that this func converts from `field.Field` into specific type?

Can we use type assertion like this:

```
        // ...
	track, ok := d.(*field.Track3{})
        if !ok {
           return nil
        }

        return getTrackDataString(in, track)
}

```

describe.go
```diff
@@ -43,6 +43,12 @@ func Describe(message *Message, w io.Writer) error {
 	// display the rest of all set fields
 	fields := message.GetFields()
 
+	// making filter map
+	filterMap := make(map[int]FilterFunc)
+	for _, filter := range filters {
```
alovak, line 48: can we here do following:
* if no filters are passed - use DefaultFilters
* otherwise - use passed filters

If we do so, all users will get "safe by default" output. If they don't want to filter fields and want to get unfiltered data, then let's create for them no-op filter so they can:

```
Describe(message, os.Stdout, iso8583.DoNotFilterFields)
```

adamdecaf: Sounds good to me.

field_filter.go
```diff
@@ -0,0 +1,112 @@
...
+		FilterField(36, Track3Filter),
+		FilterField(45, Track1Filter),
+		FilterField(52, PINFilter),
+		FilterField(55, EMVFilter),
+	}
+	return filters
+}
+
+var EMVFilter = func(in string, data field.Field) string {
+	if len(in) < emvFirstIndex+emvLastIndex {
```
adamdecaf, line 45: Getting the length of a string should use `utf8.RuneCountInString`

PR #208, opened 2023-01-10
add spec and Data type for EMV

alovak:
What:
* describe spec for ICC field containing EMV chip data
* provide structure to marshal/unmarshal EMV chip data


wadearnold:
This creates an error if the fields are not defined. Wouldn't it be better to parse them into Tag => Value pairs if it is unknown?

exp/emv/README.md
```diff
@@ -0,0 +1,5 @@
+# emv package
+
+The `emv` package is an experimental package for handling EMV data. It may have limitations and not suitable for all use-cases. You should carefully evaluate if it fits your needs, and don't hesitate in reaching out if you have any questions or concerns.
+
+If you encounter any bugs or issues, or have feedback on your experience using the package, please let us know. We welcome contributions to improve the package and if you are interesting in helping out, feel free to submit pull requests or [open new issue](https://github.com/moov-io/iso8583/issues/new).
```
adamdecaf, line 5:
```suggestion
If you encounter any bugs or issues, or have feedback on your experience using the package, please let us know. We welcome contributions to improve the package and if you are interesting in helping out, feel free to submit pull requests or [open new issue](https://github.com/moov-io/iso8583/issues/new). We hang out in the `#iso8583` channel on the [Moov Community Slack](https://slack.moov.io). 
```

adamdecaf (approved):

sort/strings.go
```diff
@@ -41,7 +41,7 @@ func StringsByHex(x []string) {
 	sort.Slice(x, func(i, j int) bool {
 		valI, err := encoding.ASCIIHexToBytes.Encode([]byte(x[i]))
 		if err != nil {
-			panic(fmt.Sprintf("failed to sort strings by hex: %v", err))
+			panic(fmt.Sprintf("failed to encode ascii hex %s to bytes : %v", x[i], err))
```
adamdecaf, line 44: This data would have been validated before sorting, so this panic isn't likely?

adamdecaf: We could return an `error` and still sort the slice in-place.

alovak: It will panic if we put an invalid hex value into the spec. So, on the first pack/unpack, it will panic. I don't think we have any other place where we validate tags.

Issue #171, opened 2022-03-11
Allow to use bitmaps with arbitrary size


alovak:
In one of our cases, I found that we need a bitmap with 4 bytes. The current implementation of the `field.Bitmap` assumes that 8 bytes is the minimal bitmap size. We have to make it default, but allow us to specify a smaller size if needed.

There will probably be no change to API, as we already specify `Length` in the `field.Spec`, we just have to use it instead of hardcoded `minBitmapLenght`.

alovak:
closed in https://github.com/moov-io/iso8583/pull/211

PR #211, opened 2023-01-31
Make bitmap of arbitrary size

alovak:
## What
* adding `DisableAutoExpand` (bool) to the field spec. Used by the `field.Bitmap`
* if DisableAutoExpand is set to true, then `field.Bitmap` will be just a fixed bitmap
* remove min and max bitmap size - now you can create bitmaps with arbitrary length (just set `DisableAutoExpand` to true)

Example of how to create fixed bitmap:
```go
	bitmap := NewBitmap(&Spec{
		Length:            2, // 2 bytes - 16 bits
		Description:       "Bitmap",
		Enc:               encoding.BytesToASCIIHex,
		Pref:              prefix.Hex.Fixed,
		DisableAutoExpand: true,
	})

	// when setting bits inside of bitmap range
	bitmap.Set(1)
	bitmap.Set(16)
        bitmap.Set(17) // is ignored as 17 is out of range
```

## Why
* https://github.com/moov-io/iso8583/issues/171
* #182

P.S. **Note**, for most card brands, the default _single_ bitmap size is 8 bytes (64 bits). Because of the previous implementation, the Length parameter was ignored and instead underlying storage for the 3 (max allowed) bitmaps was used. For all operations, it calculated the number of bitmaps based on the bits set. So, if for auto-expanding bitmaps you use 16 or something else, then bitmap will be expanded by 16 bytes for new bitmaps. That's not what you want. Just change from 16 to 8.


adamdecaf:
It sounds like `AutoExpand: false` is the current behavior so that should be the default. Changing this into an auto-expanding bitmap (by default) will break existing code.

mfdeveloper508:
LGTM
the approach is good

my advice is 
- remove DisableAutoExpand option
- AutoExpand is active if there isn't any length setting 
- AutoExpand is inactive if there is any length setting

mfdeveloper508 (approved):

alovak:
@adamdecaf our current behaviour is `AutoExpand: true` that's why I had to introduce option that is false by default (negative naming).

field/bitmap.go
```diff
@@ -34,82 +38,74 @@ func (f *Bitmap) SetSpec(spec *Spec) {
...
 		read += readDecoded
+		f.data = append(f.data, decoded...)
 
-		rawBitmap = append(rawBitmap, decoded...)
-		bitmap := utils.NewBitmapFromData(decoded)
-
-		// if no more bitmaps, exit loop
-		if !bitmap.IsSet(1) {
+		// if it's a fixed bitmap or first bit of the decoded bitmap is not set, exit loop
+		if f.spec.DisableAutoExpand || decoded[0]&(1<<7) == 0 {
```
alovak, line 104: should I use `128` instead of `1<<7` as I mix these both approaches in the code? Want to be more consistent.

adamdecaf:
Thanks for making it backwards compatible.

adamdecaf (approved):

PR #212, opened 2023-02-01
Fix binaryVarPrefixer to work with more bytes than prefix

FacundoMora:
Adjusts the length decoding, the previous implementation only works when the data provided only has the length bytes, being the real case scenario when data always haves more data than the prefix bytes.

To adjust this issue was necessary to cut the amount of bytes corresponding to the digits attribute from the data provided.


adamdecaf (approved):

adamdecaf:
Thanks for fixing this @FacundoMora!

alovak (approved):
@FacundoMora thanks a lot for the fix!

PR #219, opened 2023-03-22
Add flag to skip unexpected TLV tags

milandaji:
This PR adds a flag `SkipUnknownTLVTags` in the tag specification to enable the skip of unknown tags in TLV-coded fields.
The current implementation throws an error when receives a tag that is not in the spec.

That tag will only work in conjunction with BerTLVTag encoding.
By default the flag is off, maintaining the current behavior. The users can activate it based on their needs.

### DIFF
--- a/field/composite.go
+++ b/field/composite.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/moov-io/iso8583/encoding"
+	"github.com/moov-io/iso8583/prefix"
 	"reflect"
 	"regexp"
 
@@ -325,7 +327,7 @@ func (f *Composite) UnmarshalJSON(b []byte) error {
 	}
 
 	for tag, rawMsg := range data {
-		if _, ok := f.spec.Subfields[tag]; !ok {
+		if _, ok := f.spec.Subfields[tag]; !ok && !skipUnknownTLVTags(f) {
 			return fmt.Errorf("failed to unmarshal subfield %v: received subfield not defined in spec", tag)
 		}
 
@@ -424,12 +426,20 @@ func (f *Composite) unpackSubfieldsByTag(data []byte) (int, error) {
 			tagBytes = f.spec.Tag.Pad.Unpad(tagBytes)
 		}
 		tag := string(tagBytes)
-		if _, ok := f.spec.Subfields[tag]; !ok {
+		if _, ok := f.spec.Subfields[tag]; !ok && !skipUnknownTLVTags(f) {
 			return 0, fmt.Errorf("failed to unpack subfield %v: field not defined in Spec", tag)
 		}
 
 		field, ok := f.subfields[tag]
 		if !ok {
+			// Obtain the length of the unknown tag and add it to the offset.
+			if skipUnknownTLVTags(f) {
+				fieldLength, readed, err := prefix.BerTLV.DecodeLength(999, data[offset:])
+				if err != nil {
+					return 0, err
+				}
+				offset += fieldLength + readed
+			}
 			continue
 		}
 
@@ -490,4 +500,8 @@ func getFieldIndexOrTag(field reflect.StructField) (string, error) {
 	}
 
 	return "", nil
+}
+
+func skipUnknownTLVTags(field *Composite) bool {
+	return field.spec.Tag != nil && field.spec.Tag.SkipUnknownTLVTags && field.spec.Tag.Enc == encoding.BerTLVTag
 }
--- a/field/composite.go
+++ b/field/composite.go
@@ -327,7 +327,7 @@ func (f *Composite) UnmarshalJSON(b []byte) error {
 	}
 
 	for tag, rawMsg := range data {
-		if _, ok := f.spec.Subfields[tag]; !ok && !skipUnknownTLVTags(f) {
+		if _, ok := f.spec.Subfields[tag]; !ok && !f.skipUnknownTLVTags() {
 			return fmt.Errorf("failed to unmarshal subfield %v: received subfield not defined in spec", tag)
 		}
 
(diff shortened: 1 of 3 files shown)


field/composite.go
```diff
@@ -491,3 +501,7 @@ func getFieldIndexOrTag(field reflect.StructField) (string, error) {
 
 	return "", nil
 }
+
+func skipUnknownTLVTags(field *Composite) bool {
+	return field.spec.Tag != nil && field.spec.Tag.SkipUnknownTLVTags && field.spec.Tag.Enc == encoding.BerTLVTag
+}
```
alovak, line 507:
```suggestion
func (f *Composite) shouldSkipUnknownTLVTags() bool {
	return f.spec.Tag != nil && f.spec.Tag.SkipUnknownTLVTags && f.spec.Tag.Enc == encoding.BerTLVTag
}
```

field/composite.go
```diff
@@ -424,12 +426,20 @@ func (f *Composite) unpackSubfieldsByTag(data []byte) (int, error) {
 			tagBytes = f.spec.Tag.Pad.Unpad(tagBytes)
 		}
 		tag := string(tagBytes)
-		if _, ok := f.spec.Subfields[tag]; !ok {
+		if _, ok := f.spec.Subfields[tag]; !ok && !skipUnknownTLVTags(f) {
```
alovak, line 429: I would move the code that changes `offset` from the block below into this `if` block. If `tag` is not defined in the spec, it will not be in the `f.subfields[tag]`. It means we can either error or skip it right here (in this block) and `continue` without calling the code below.

field/composite_test.go
```diff
@@ -335,6 +335,28 @@ func TestTLVPacking(t *testing.T) {
...
+	t.Run("Unpack correctly deserialises bytes to the data struct skipping unexpected tags", func(t *testing.T) {
+		// Turn on the skipping unexpected tags capability and turn it off at the end of test
+		tlvTestSpec.Tag.SkipUnknownTLVTags = true
+		defer func() {
+			tlvTestSpec.Tag.SkipUnknownTLVTags = false
+		}()
+
+		composite := NewComposite(tlvTestSpec)
+
+		// Data contains tags 9F36, 9A, 9F02 and 9F37
```
alovak, line 347: I would like to highlight the tags that are not in the spec

field/composite_test.go
```diff
@@ -399,6 +421,15 @@ func TestTLVPacking(t *testing.T) {
 		require.Equal(t, "027F", data.F9F36.Value())
 		require.Equal(t, "047F", data.F9F3B.F9F45.Value())
 	})
+
+	t.Run("Unpack throws an error due unexpected tags", func(t *testing.T) {
+		composite := NewComposite(tlvTestSpec)
+
+		// Data contains tags 9F36, 9A, 9F02 and 9F37
```
alovak, line 428: I would like to highlight the tags that are not in the spec

field/composite_test.go
```diff
@@ -1117,6 +1148,61 @@ func TestCompositeJSONConversion(t *testing.T) {
...
+		require.JSONEq(t, json, string(actual))
+	})
+
+	t.Run("UnmarshalJSON TLV data skipping unexpected tags", func(t *testing.T) {
+		// Turn on the skipping unexpected tags capability and turn it off at the end of test
+		tlvTestSpec.Tag.SkipUnknownTLVTags = true
+		defer func() {
+			tlvTestSpec.Tag.SkipUnknownTLVTags = false
+		}()
+		json_extra_tags := `{"9A":"210720","9F02":"000000000501", "9F37": "9badbcab"}`
```
alovak, line 1175: I would like to highlight the tags that are not in the spec

alovak:
@milandaji that's amazing PR! Thank you for a great job!

Let's address some tiny things and then we can merge it.

field/composite.go
```diff
@@ -424,12 +426,20 @@ func (f *Composite) unpackSubfieldsByTag(data []byte) (int, error) {
...
 		field, ok := f.subfields[tag]
 		if !ok {
+			// Obtain the length of the unknown tag and add it to the offset.
+			if skipUnknownTLVTags(f) {
+				fieldLength, readed, err := prefix.BerTLV.DecodeLength(999, data[offset:])
+				if err != nil {
+					return 0, err
+				}
+				offset += fieldLength + readed
+			}
```
alovak, line 442: I mean move this code into the previous `if` block.

field/composite.go
```diff
@@ -424,12 +426,20 @@ func (f *Composite) unpackSubfieldsByTag(data []byte) (int, error) {
...
-		if _, ok := f.spec.Subfields[tag]; !ok {
+		if _, ok := f.spec.Subfields[tag]; !ok && !skipUnknownTLVTags(f) {
 			return 0, fmt.Errorf("failed to unpack subfield %v: field not defined in Spec", tag)
 		}
 
 		field, ok := f.subfields[tag]
 		if !ok {
+			// Obtain the length of the unknown tag and add it to the offset.
+			if skipUnknownTLVTags(f) {
+				fieldLength, readed, err := prefix.BerTLV.DecodeLength(999, data[offset:])
```
alovak, line 437: First I wanted to ask if 999 is the max length, then I found that it's not used inside berTLV as length is read from data. I suggest to:
```suggestion
				fieldLength, readed, err := prefix.BerTLV.DecodeLength(ignoredMaxLen, data[offset:])
```

and define `ignoredMaxLen` before the method:

```go
// Because BER-TLV lengths are decoded dynamically, the maxLen method
// argument is ignored.
const ignoredMaxLen int = 0
```

milandaji: Great!
I didn't realize that maxLength is being ignored in `BerTLV` decoder. Now is implemented as you suggested.

field/composite.go
milandaji: Done.
Thank you for the suggestion.

milandaji:
@alovak Hi!
All the suggestions were addressed :D

field/composite.go
```diff
@@ -425,6 +427,18 @@ func (f *Composite) unpackSubfieldsByTag(data []byte) (int, error) {
 		}
 		tag := string(tagBytes)
 		if _, ok := f.spec.Subfields[tag]; !ok {
+			if f.skipUnknownTLVTags() {
+				// Obtain the length of the unknown tag and add it to the offset.
+				// Because BER-TLV lengths are decoded dynamically, the maxLen method argument is ignored.
+				const ignoredMaxLen int = 0
```
alovak, line 433: let's move definition of `ignoredMaxLen` out of the method (it can be line 415) so we don't define const for each unpack.

alovak:
I’ll give it one last review tomorrow and I believe we can merge and release the change.

alovak (approved):
@milandaji thanks a lot!

Issue #220, opened 2023-04-01
prefix.BCD.Fixed doesn't calculate the encoding length correctly


ddvk:
When using BCD.Fixed, my expectation was that the length is in Nibbles i.e. number of digits.

However, encoding checks and uses the length in bytes, but the Decoding uses the length in digits.

There is no error during packing, and the other side cannot  correctly decode the message as the field is larger than the spec and further fields are clobbered.

Example
```go
package main

import (
	"encoding/hex"
	"testing"

	"github.com/moov-io/iso8583"
	"github.com/moov-io/iso8583/encoding"
	"github.com/moov-io/iso8583/field"
	"github.com/moov-io/iso8583/padding"
	"github.com/moov-io/iso8583/prefix"
	"github.com/stretchr/testify/require"
)

var spec = &iso8583.MessageSpec{
	Fields: map[int]field.Field{
		0: field.NewNumeric(&field.Spec{
			Length:      4,
			Description: "Message Type Indicator",
			Enc:         encoding.BCD,
			Pref:        prefix.BCD.Fixed,
		}),
		1: field.NewBitmap(&field.Spec{
			Description: "Bitmap",
			Enc:         encoding.Binary,
			Pref:        prefix.Binary.Fixed,
		}),
		2: field.NewNumeric(&field.Spec{
			Length:      2,
			Description: "SomeFixedField",
			Enc:         encoding.BCD,
			Pref:        prefix.BCD.Fixed,
			Pad:         padding.Left('0'),
		}),
		3: field.NewNumeric(&field.Spec{
			Length:      3,
			Description: "SomeVarField",
			Enc:         encoding.BCD,
			Pref:        prefix.BCD.LLLL,
		}),
	},
}

func TestFixed(t *testing.T) {
	msg := iso8583.NewMessage(spec)
	msg.MTI("1234")
	field2 := "1234"
	msg.Field(2, field2)
	out, err := msg.Pack()
	require.NoError(t, err)
	t.Log(hex.EncodeToString(out))
	in := iso8583.NewMessage(spec)
	err = in.Unpack(out)
	require.NoError(t, err)
	result, _ := in.GetField(2).String()
	require.Equal(t, field2, result)
}

func TestVariable(t *testing.T) {
	msg := iso8583.NewMessage(spec)
	msg.MTI("1234")
	field3 := "456"
	msg.Field(3, field3)
	out, err := msg.Pack()
	require.NoError(t, err)
	t.Log(hex.EncodeToString(out))
[...]

Issue #58, opened 2021-05-24
field constraints support


vtolstov:
some fields not only must have iso8583 types, but also must conform to some other stuff like len of field, or number of spaces or digits , may be cool to add Validate func(field.Field) error to spec?
So after we fill the fields and Validate func not nil we can call it and get reasonable error if it present?
Also we can create predefined funcs as example for date time fields for example...
i think that this will be cool

alovak:
@vtolstov it you have a chance to make a draft/proposal for this that would be really helpful.

We can add validation to each field (Numeric, String, Datetime, etc.) or we can add Validation to the Spec. Maybe for the spec we can add something like a format that will be passed to the field validator. What are your thoughts?

PR #221, opened 2023-04-07
fix for #220 bcd length encoding

ddvk:
Here's is my proposition for fixing the bcd length encoding. This also fixes the variable encoding where an odd length was encoded with the next even number.

The problem I had with bcd is that there is no way to find out whether the length of original data was odd or even, that's why the `Encoder interface, Encode` return type change.

I couldn't come up with a more elegant solution.

### DIFF
--- a/encoding/ascii.go
+++ b/encoding/ascii.go
@@ -13,16 +13,16 @@ var (
 
 type asciiEncoder struct{}
 
-func (e asciiEncoder) Encode(data []byte) ([]byte, error) {
+func (e asciiEncoder) Encode(data []byte) ([]byte, int, error) {
 	var out []byte
 	for _, r := range data {
 		if r > 127 {
-			return nil, utils.NewSafeError(fmt.Errorf("invalid ASCII char: '%s'", string(r)), "failed to perform ASCII encoding")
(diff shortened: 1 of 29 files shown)


adamdecaf:
Thanks for opening this PR! Can you update the godoc comments to mention what that integer represents?

ddvk:
the godoc of the Encode method?

adamdecaf:
cc @alovak 

Can we fix and amend that typo? @ddvk

adamdecaf:
cc @alovak 

Can we fix and amend that typo? @ddvk

alovak (changes requested):
I'm not sure that there is an issue with BCD encoding and prefixes. In the issue #220 specified length does not match the length of the value that is set to the field. Here is my comment about it: https://github.com/moov-io/iso8583/issues/220#issuecomment-1516072707

Maybe, we should add length validation to avoid such issues (https://github.com/moov-io/iso8583/issues/58).

alovak:
@ddvk thanks for raising the issue and your contributions. You helped us to realize that there is an issue with field packing/unpacking.  I'm closing this PR as the issue is going to be addressed by this PR: https://github.com/moov-io/iso8583/pull/227

ddvk:
sure, thanks a lot.

PR #226, opened 2023-04-21
fix typo

SDTakeuchi:
I discovered a small typographical error in the `encoding\bertlv_test.go` file and have corrected it by changing `3nd` to `3rd`.


adamdecaf (approved):

adamdecaf:
Thanks for the fix!

PR #229, opened 2023-05-02
Composites with private bitmaps

FacundoMora:
Hi Everyone!,

A few month ago me and my team encountered with a peculiar type of field, that implied a structure like this:
_[LL BCD Length] + [3 bytes bitmap] + [Subfields]_

In this structure we were missing two features: 
- **Arbitrary sized bitmaps**, when defining it on spec we could indicate its block size (3 bytes) and the amount of blocks.
- **Some way to have a field with subfields, where its subfields existence could be determined by a private bitmap** (We are talking about you, composite 🤨)

As you already know, in a previous release, a change about arbitrary size bitmaps made by @alovak took place, **in this changes our first feature needed to solve this situation was finished**.

So now, we bring the second feature, where we can use this bitmap as a private bitmap of a composite 👏 

To achieve this we decided to reutilize most of the logic in iso8583.Message, specifically the way that uses its bitmap to find the message fields for packing and unpacking. Obviously we had to modify a bit of code to make sense on a composite context.

PD: Yes, **iso8583.Message could be replaced as a composite** given that their behaviour are pretty the same, but with @alovak agree on leave it for a future potential change, so that these changes do not become of great impact.

Finally, a minor change we made was that we moved the spec validation location inside the SetSpec along with the ContructSubfields call. We made this because we thought this makes more sense given that this two things have to run if the spec is changed.

I look forward for your comments on this, thanks!


alovak:
@FacundoMora hey! Thanks for the PR! Great job! I'm reviewing it.

field/composite.go
```diff
@@ -113,11 +118,35 @@ func (f *Composite) getSubfields() map[string]Field {
 // should only pass None or nil values for ths type. Passing any other value
 // will result in a panic.
 func (f *Composite) SetSpec(spec *Spec) {
```
adamdecaf, line 120: Can we return an error here instead of the panics? It breaks the public API, but returning valid errors is more important to me.

alovak (commented):
Here are some of my comments:
1. It seems if we add a separate field `Bitmap` into the Spec as we did for `Tag` that should simplify the code a little bit.
1. I think we can add a bitmap to the composite field without changing the **structure** of the constructor and `SetSpec`. Let’s keep them as they are. I'm not sure that such change is needed.
2. In the validation we can have distinct blocks for validation when Bitmap is set and when … to make it simpler. In the PR I can’t easily understand all conditions 🙈 
4. Let’s use []string and not []int, we still need to convert int to string or string to int, but keeping it []string allows us to use the same sort package that we already have and use
5. I would suggest adding validation of all indexes (that they can be converted into `int` without errors) into `validateCompositeSpec` when `Bitmap` is set.

field/composite.go
FacundoMora: We initially thought on that, but the problem with this is that we have to change the field.Field interface, this means that we will have to change all the implementations for this interface, what do you think?

alovak: I have checked, and it is not an easy change to make because `SetSpec` is being used somewhere along the call chain by `func NewMessage(spec *MessageSpec) *Message` - and I don't know how big would be the breaking change if we add `error` to `NewMessage`.

field/composite.go
```diff
@@ -295,6 +307,12 @@ func (f *Composite) Bytes() ([]byte, error) {
 	return f.pack()
 }
 
+// Bitmap returns the parsed bitmap instantiated on the key "0" of the spec.
+// In case the bitmap is not instantiated on the spec, returns nil.
+func (f *Composite) Bitmap() *Bitmap {
+	return f.spec.Bitmap
```
alovak, line 313: should it be `f.Bitmap`?

FacundoMora: Done, I made a deep copy of spec.Bitmap and saved it to f.Bitmap in case f.Bitmap is nil, is this ok?

field/composite.go
```diff
@@ -413,6 +498,44 @@ func (f *Composite) unpackSubfields(data []byte, isVariableLength bool) (int, er
...
+	read, err := f.Bitmap().Unpack(data[off:])
+	if err != nil {
+		return 0, fmt.Errorf("failed to unpack bitmap: %w", err)
+	}
+
+	off += read
+
+	for i := 1; i <= f.Bitmap().Len(); i++ {
+		if f.Bitmap().IsSet(i) {
+			iStr := fmt.Sprintf("%v", i)
```
alovak, line 518:
```suggestion
			iStr := strconv.Itoa(i)
```

field/composite.go
```diff
@@ -347,6 +381,54 @@ func (f *Composite) UnmarshalJSON(b []byte) error {
...
+	// pack fields
+	for _, i := range f.orderedSpecFieldTags {
+		// If this ordered field is not set, continue to the next field.
+		if _, ok := f.setSubfields[i]; !ok {
+			continue
+		}
+
+		field, ok := f.subfields[i]
+		if !ok {
+			return nil, fmt.Errorf("failed to pack subfield %v: no specification found", i)
```
alovak, line 419:
```suggestion
			return nil, fmt.Errorf("failed to pack subfield %s: no specification found", i)
```

field/composite.go
```diff
@@ -347,6 +381,54 @@ func (f *Composite) UnmarshalJSON(b []byte) error {
...
+		}
+		packedField, err := field.Pack()
+		if err != nil {
+			return nil, fmt.Errorf("failed to pack subfield %v (%s): %w", i, field.Spec().Description, err)
```
alovak, line 423:
```suggestion
			return nil, fmt.Errorf("failed to pack subfield %s (%s): %w", i, field.Spec().Description, err)
```

field/composite.go
```diff
@@ -413,6 +498,44 @@ func (f *Composite) unpackSubfields(data []byte, isVariableLength bool) (int, er
...
+
+	off += read
+
+	for i := 1; i <= f.Bitmap().Len(); i++ {
+		if f.Bitmap().IsSet(i) {
+			iStr := strconv.Itoa(i)
+
+			fl, ok := f.subfields[iStr]
+			if !ok {
+				return 0, fmt.Errorf("failed to unpack subfield %v: no specification found", iStr)
```
alovak, line 522:
```suggestion
				return 0, fmt.Errorf("failed to unpack subfield %s: no specification found", iStr)
```

field/composite.go
```diff
@@ -413,6 +498,44 @@ func (f *Composite) unpackSubfields(data []byte, isVariableLength bool) (int, er
...
+			}
+
+			read, err = fl.Unpack(data[off:])
+			if err != nil {
+				return 0, fmt.Errorf("failed to unpack subfield %v (%s): %w", iStr, fl.Spec().Description, err)
```
alovak, line 527:
```suggestion
				return 0, fmt.Errorf("failed to unpack subfield %s (%s): %w", iStr, fl.Spec().Description, err)
```

field/composite.go
```diff
@@ -467,26 +590,47 @@ func (f *Composite) skipUnknownTLVTags() bool {
...
 	}
-	if spec.Tag != nil && spec.Tag.Enc == nil && spec.Tag.Length > 0 {
-		return fmt.Errorf("Composite spec requires a Tag.Enc to be defined if Tag.Length > 0")
+
+	// If bitmap is defined, validates subfields keys.
+	// spec.Tag is not validated.
+	if spec.Bitmap != nil {
+		for key := range spec.Subfields {
+			parsedKey, err := strconv.Atoi(key)
+
```
alovak, line 608:
```suggestion
```

field/composite.go
```diff
@@ -467,26 +590,47 @@ func (f *Composite) skipUnknownTLVTags() bool {
...
+	if spec.Tag.Enc == nil && spec.Tag.Length > 0 {
+		return fmt.Errorf("Composite spec requires a Tag.Enc to be defined if Tag.Length > 0")
+	}
+
 	return nil
 }
 
 func orderedKeys(kvs map[string]Field, sorter sort.StringSlice) []string {
-	keys := make([]string, 0)
+	keys := make([]string, 0, len(kvs))
```
alovak, line 633: nice!

field/composite_test.go
```diff
@@ -95,6 +95,96 @@ var (
...
 
+	compositeTestSpecWithDefaultBitmap = &Spec{
+		Length:      32,
+		Description: "Test Spec",
+		Pref:        prefix.ASCII.LL,
+		Bitmap: NewBitmap(&Spec{
+			Length:      8,
+			Description: "Bitmap",
+			Enc:         encoding.BytesToASCIIHex,
+			Pref:        prefix.Hex.Fixed,
```
alovak, line 106: I think you should add `DisableAutoExpand: true` here, otherwise, if you set subfield `1`, then the packed length of the bitmap will be 16 bites. You can run an experiment to disprove this :)

But it's 100% that you don't want bitmap to auto expand like it does in the message.

field/composite_test.go
```diff
@@ -95,6 +95,96 @@ var (
...
+			}),
+		},
+	}
+
+	compositeTestSpecWithSizedBitmap = &Spec{
+		Length:      30,
+		Description: "Test Spec",
+		Pref:        prefix.ASCII.LL,
+		Bitmap: NewBitmap(&Spec{
+			Length:      3,
```
alovak, line 148: Let's add `DisableAutoExpand: true` to avoid bitmap expanding

field/composite_test.go
```diff
@@ -952,6 +1042,309 @@ func TestCompositePackingWithTags(t *testing.T) {
...
+
+		composite := NewComposite(compositeTestSpecWithSizedBitmap)
+		err := composite.SetData(data)
+		require.NoError(t, err)
+
+		// F2 fails to unpack - it requires length to be defined instead of AB.
+		read, err := composite.Unpack([]byte("16602000AB060102YZ"))
+		require.Equal(t, 0, read)
+		require.Error(t, err)
+		require.EqualError(t, err, "failed to unpack subfield 2 (String Field): failed to decode length: strconv.Atoi: parsing \"AB\": invalid syntax")
```
alovak, line 1281: I think that in this case we potentially can leak sensitive data and I suggest using `SafeError` here. You can find examples of how we use it in the iso858 package.

alovak:
@FacundoMora It looks good, just some tiny changes are needed and we can merge it.

field/composite.go
FacundoMora: Done

field/composite.go
FacundoMora: Done

field/composite.go
FacundoMora: Done

field/composite.go
FacundoMora: Done

field/composite.go
FacundoMora: Thanks! 😄

field/composite_test.go
FacundoMora: Done

field/composite_test.go
alovak: https://github.com/moov-io/iso8583/blob/49dc79f0a98c23673bc1baaabb3fd1d986175ece/encoding/ascii.go#L20

but I've checked that in your case it's not needed. We convert index/tag which is expected to be number and is in the spec only, not in the message data.

field/bitmap.go
```diff
@@ -200,6 +200,10 @@ func (f *Bitmap) IsSet(n int) bool {
 		return false
 	}
 
+	if n == 1 && !f.spec.DisableAutoExpand {
```
alovak, line 203: Could you please elaborate on the rationale behind this specific change? I'm having difficulty understanding the problem they're intended to solve.

FacundoMora: The issue locates on the pack func, when bitmap is auto expanded (meaning that the first bit is active and DisableAutoExpand is false) this will still try to pack the field 1. 
To fix this issue I added this validation, to return false when field 1 is active but not really because of the auto expansion. This condition doesn't contemplates the bits 65, 129 and on, yet.

FacundoMora: The final decision for this was to not allow composite specs with bitmaps with auto expansion, with this rule this problem is solved for now.

alovak (approved):
Great job!

PR #230, opened 2023-05-20
add a method to remove previously set fields from a message

ddvk:
I have the following use case: I clone an incoming message, then modify some fields and send it back. Due to the spec, some fields should not be send back, so this allows me to remove those fields.

### DIFF
--- a/message.go
+++ b/message.go
@@ -84,6 +84,15 @@ func (m *Message) Field(id int, val string) error {
 	return fmt.Errorf("failed to set field %d. ID does not exist", id)
 }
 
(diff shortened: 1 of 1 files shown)


alovak:
Hey, @ddvk please check this one: https://github.com/moov-io/iso8583/pull/113

in summary it’s not safe to create copy of the message and in the PR comments you can find a suggested way how to copy data you need from one message into another using Unmarshal and Marshal

ddvk:
ok, thanks, it's just that there are like 50 fields inthe message and i unmarshal only the ones i needed

PR #236, opened 2023-06-05
Encoding.ascii tlv tag

jerome-laforge:
Hello,
I want to be able to ignore unknown subfield with ASCII Encoder.
This PR tries to add this feature for ASCII Encoder (as `BerTLVTag`).

Can you review this PR?

Thank in advance,


alovak:
With the suggested change with LLL encoders, you still may get an "unknown" field with length that doesn't match specified in the Tag. Let's say you assume that all fields have L, but what if some will have LL?

With BerTLV it's clear how we can skip unknown fields - their length is encoded into data, not defined in the spec. But with regular fields, I don't see how it's possible.

alovak:
But how do you know that they don't add LLL field when you expect it to be L?

jerome-laforge:
> But how do you know that they don't add LLL field when you expect it to be L?

Unfortunately, I can't. 
If they add LLL instead of L, then it is likely that the decoding will fail.
I can understand if you think that it is better to reject this PR because this change doesn't expect the quality or so on.
I will keep our driver on my fork :)

alovak:
I have one option that may work here without too much changes.

Here it is:
* Add `Pref` into `TagSpec` - with comment that by default it's prefix.BerTLV otherwise, it will be used for all unknown tags and may fail to parse if unknown tag prefixer will be different
* update `unpackSubfieldsByTag` in the following way:
```go
			if f.skipUnknownTLVTags() {
				var pref prefix.Prefixer
				if f.Pref != nil {
					pref = f.Pref
				} else {
					pref = prefix.BerTLV
				}

				// Obtain the length of the unknown tag and add it to the offset.
				// Because BER-TLV lengths are decoded dynamically, the maxLen method argument is ignored.
				fieldLength, readed, err := pref.DecodeLength(ignoredMaxLen, data[offset:])
				if err != nil {
					return 0, err
				}
				offset += fieldLength + readed
				continue
			}
```

3. Update `skipUnknownTLVTags` like this:
```go
func (f *CustomCompositeField) skipUnknownTLVTags() bool {
	return f.spec.Tag != nil && f.spec.Tag.SkipUnknownTLVTags && (f.spec.Tag.Enc == encoding.BerTLVTag || f.Pref != nil)
}
```

In this way we:
* don't add new encodings
* don't break API as we add new field and don't change current behaviour
* support your case


What do you think about it?

jerome-laforge:
> What do you think about it?

It seems great.

But, could you also show me the code of `CustomCompositeField` struct?

alovak:
@jerome-laforge `CustomCompositeField` was just my internal experiment I decided to run to see if you can create a custom field as an alternative to this PR. Then I realized that we can make the change I suggested and it is acceptable (not introducing a lot of new things or changing a lot) :)

jerome-laforge:
By `f.Pref`, did you mean `f.spec.Pref`?
If yes, then I am surely miss something, because `f.spec.Pref` is for field (ex: field 38), isn't it?

For example, I have field 38 on LLL with 999 maxLen, but if all partners decide that all its sub-fields are encode on L.
Did I miss something?

jerome-laforge:
or maybe, you want enforce that field & sub-fields have the same encoding?

alovak:
Add `Pref prefixer.Prefixer` to `field.TagSpec`:

```go
type TagSpec struct {
	Pref prefixer.Prefixer
}
```

Sorry, my example is from the experiment code and it says `f.Pref`. I'll update it.

alovak:
We are not adding new encodings, we can add length prefixer for the `TagSpec` and use it if we want to skip unknown tags and it's not BerTLV.

jerome-laforge:
Please find the update.
Let me know if something (for example naming or doc is not ok for you).

Thank again for your support

field/composite.go
```diff
@@ -558,9 +563,22 @@ func (f *Composite) unpackSubfieldsByTag(data []byte) (int, error) {
...
 			if f.skipUnknownTLVTags() {
-				// Obtain the length of the unknown tag and add it to the offset.
-				// Because BER-TLV lengths are decoded dynamically, the maxLen method argument is ignored.
-				fieldLength, readed, err := prefix.BerTLV.DecodeLength(ignoredMaxLen, data[offset:])
+				var (
+					pref   prefix.Prefixer
+					maxLen int
+				)
+				// to obtain the length of the unknown tag and add it to the offset we need to decode its length
+				if f.spec.Tag.Enc != encoding.BerTLVTag {
```
alovak, line 571: What if `f.spec.Tag.PrefUnknownTLV` is `nil`?

To keep it backward compatible, I would rely on the `if f.spec.Tag.PrefUnknownTLV != nil { ... } else { use as it was before } `

jerome-laforge: It is not possible that `f.spec.Tag.Enc != encoding.BerTLVTag && f.spec.Tag.PrefUnknownTLV == nil`)  as we previously check with:
```go
func (f *Composite) skipUnknownTLVTags() bool {
	return f.spec.Tag != nil && f.spec.Tag.SkipUnknownTLVTags && (f.spec.Tag.Enc == encoding.BerTLVTag || f.spec.Tag.PrefUnknownTLV != nil)
}
```

Please find this update, is it suitable for you?

jerome-laforge:
> @jerome-laforge it looks good! One final question. May I ask you to keep tests only for one type of prefix (you can choose any L or LL or LLL)? I'm not sure there is a benefit in testing composite with unknown tags all prefixers as they have their own tests.

Ok, I will do that

alovak (approved):
Great job @jerome-laforge ! Thanks a lot for the contributions!

PR #237, opened 2023-06-06
Display composite field subfields in iso8583.Describe

alovak:
What:
* display composite field subfields when dump message with `iso8583.Describe`
* split bitmap bits in 32 bits long lines and add bits indexes to make it easier to see which bit is set

Here is the new output (filtered):

```
ISO 8583 Message:
MTI..........: 0800
Bitmap HEX...: 82200000000000020C00000020000000
Bitmap bits..:
[1-8]10000010 [9-16]00100000 [17-24]00000000 [25-32]00000000
[33-40]00000000 [41-48]00000000 [49-56]00000000 [57-64]00000010
[65-72]00001100 [73-80]00000000 [81-88]00000000 [89-96]00000000
[97-104]00100000 [105-112]00000000 [113-120]00000000 [121-128]00000000
F0   Message Type Indicator......: 0800
F7   Transmission Date and Time..: 0606114142
F11  System Trace Audit Number...: 123456
F63 Private Fields SUBFIELDS:
-------------------------------------------
Bitmap HEX...: 800000
Bitmap bits..:
[1-8]10000000 [9-16]00000000 [17-24]00000000
F1   Institution ID..: 0002
------------------------------------------
F69  Institution Country Code............................: 840
F70  Management Information Code..................: 301
F99  Institution Identification Code....................: 12345678901
```

**It's a breaking change:**
* output format of the CLI and `iso8583.Describe` is changed
* field filters work with a `string` field id or tag instead of an `int`


alovak:
@wadearnold I was thinking about adding more colors ;)

describe.go
```diff
@@ -92,13 +152,61 @@ func Describe(message *Message, w io.Writer, filters ...FieldFilter) error {
...
+	keys = append(keys, nonNumericKeys...)
 
 	return keys
 }
+
+// splitAndAnnotate splits bits blocks and annotates them with bit numbers
+// and splits them by 32 bits if needed
+func splitAndAnnotate(bits string) string {
+	bitBlocks := strings.Split(bits, " ")
+	if len(bitBlocks) == 0 {
```
adamkornafeld, line 188:
```suggestion
	bitBlocks := strings.Split(bits, " ")
	if len(bitBlocks) == 1 {
```

If the 2nd arg, `separator` of `Split` is not empty and the first arg `bits` does not contain the separator, the returned slice will have a length of 1 with one element being the input - according to the docs.

adamkornafeld (approved):
Looks good!

PR #247, opened 2023-06-29
Enhance Readme

bpross:
Make some small changes to the README to help future devs avoid the rabbit hole I went down.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ go get github.com/moov-io/iso8583
 
 ### Define your specification
 
-Currently, we support following ISO 8583 specifications:
+Currently, we have defined the following ISO 8583 specifications:
 
 * [Spec87ASCII](./specs/spec87ascii.go) - 1987 version of the spec with ASCII encoding
 * [Spec87Hex](./specs/spec87hex.go) - 1987 version of the spec with Hex encoding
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,7 +58,7 @@ go get github.com/moov-io/iso8583
 
 ### Define your specification
 
-Currently, we support following ISO 8583 specifications:
+Currently, we have defined the following ISO 8583 specifications:
```
bpross, line 61: This was a little confusing for myself and my team. We assumed by "support" we needed to add a new spec to support anything not based off of the 1987 spec. That's incorrect. The library provides some sample specs already, but there is support for newer versions of the spec.

README.md
```diff
@@ -124,6 +124,56 @@ spec := &iso8583.MessageSpec{
 }
 ```
 
+The following example creates a full specification with three individual fields (excluding MTI and Bitmap). It differs from the example above, by [showing the expandability of the bitmap field](https://github.com/moov-io/iso8583/blob/5b24f23a5a02206d59baa033ed040f79478b6ecb/field/bitmap.go#L94-L109). This is useful for specs that define both a primary and secondary bitmap.
```
bpross, line 127: I added this, because I was a little confused about how the library interacted with a secondary bitmap. Happy to change this around a bit, but I think we should have an example that shows the library does support a secondary bitmap.

I bring this up, because the 1993 spec has the secondary bitmap defined as field 1, which is where the library puts the primary bitmap.

PR #250, opened 2023-07-20
build: force go1.20.5

adamdecaf:


adamdecaf:
Seems to be an issue specific to go1.20.5 but fixed in go1.20.6.

PR #267, opened 2023-08-26
Extend the Header interface with Name()

### LINKED ISSUE #144: Header interface() should include Name()
The [Header interface](https://github.com/moov-io/iso8583/blob/679e5a69f86ddb79dc0e35f61a0c1ad8dcfaf064/network/header.go#L59) should include a Name() property like [Specs](https://github.com/moov-io/iso8583/blob/679e5a69f86ddb79dc0e35f61a0c1ad8dcfaf064/specs/spec87ascii.go#L13) do.  You can then switch and check which compiled header you are working with.

PumpkinSeed:
Resolve #144 

- Added the Name() method for the Header interface
- Implement the missing method for the structs implement the Header

### DIFF
--- a/network/ascii_4bytes_header.go
+++ b/network/ascii_4bytes_header.go
@@ -25,6 +25,10 @@ func (h *ASCII4BytesHeader) Length() int {
 	return h.Len
 }
 
+func (h *ASCII4BytesHeader) Name() string {
+	return "4 bytes ASCII"
+}
+
 func (h *ASCII4BytesHeader) WriteTo(w io.Writer) (int, error) {
 	return fmt.Fprintf(w, "%04d", h.Len)
 }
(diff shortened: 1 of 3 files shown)


alovak:
@PumpkinSeed May I ask you to share some context about your use case for headers and `Name`?

I'm asking because I think that we should move headers into https://github.com/moov-io/iso8583-connection -> `headers` package

PumpkinSeed:
Well, I opened the issue board and I saw the #144. So I thought I resolve this issue by adding this. That's my use-case for now. I'm just trying to understand the library, because we have a basic iso8583 package. I think the best way to understand something is via contributions.

alovak:
@PumpkinSeed thanks for the context! I had to close that ticket so no one would take it. Sorry.


this package is focused on message building and parsing. For networking, which includes headers, we have a separate repository https://github.com/moov-io/iso8583-connection. So, I want to move headers there. But it will be a breaking change... so I tried to postpone it...

PR #273, opened 2023-09-06
Experimental: simplify data structs

alovak:
The goal of this PR is to implement (now it's more like proof of concept) the proposal #277 to use go basic types in the data structs instead of `*field.String`, `*field.Numeric`, etc. like this:

```go
	type authRequestData struct {
		MTI                  string `index:"0"`
		PrimaryAccountNumber string `index:"2"`
		ProcessingCode       int    `index:"3"`
	}
```

### DIFF
--- a/exp_test.go
+++ b/exp_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/moov-io/iso8583"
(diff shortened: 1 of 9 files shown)


alovak:
this PR was a proof of concept. Now all ideas are implemented in the module.

Issue #269, opened 2023-08-30
avoid using panic in the code


alovak:
There are some areas in the module where we use panic. However, it's not best practice to employ panic within a library. It would be beneficial to explore alternatives for eliminating its use while maintaining as much backward compatibility as possible.

adamdecaf:
We can return errors where there weren't some before. It sorta breaks the API, but in a fairly clean way that people can check. I hope no one was relying on `panic` behavior.

SahibYar:
we can write wrapper functions for calling current functions using panics with `recover()` for backward compatibility in `defer` functions to handle `panics`.  Refrence: https://go.dev/blog/defer-panic-and-recover

and in the meanwhile we can start rewriting functions by returning errors or `utils.NewSafeError`

SahibYar:
@alovak we can start adding `defer` functions with `recover` calls in functions which are calling `panic` to maintain backward compatibility for now and add comment that this function will be `Deprecated` in coming releases instead use new function returning `error` or `utils.NewSafeError`. 

```go
defer func() {
    if r := recover(); r != nil {
	fmt.Println("Recovered error:", r)
   }
}()
```

Or the other way is start writing functions straight away as @adamdecaf mentioned in their comment.

Please suggest the acceptable approach, and I will raise a PR or may be extend @PumpkinSeed work.

PumpkinSeed:
@SahibYar I have a question if you don't mind, because I don't understand certain parts of that. [I quickly created a showcase based on your recover suggestion](https://go.dev/play/p/HLh3NFqfxDl), please update the code since I'm sure it's not matching your expectations.

I assume someone relies on the panic's behavior by using recover function. If we catch the panic before it hits the user of the library's recover, how it supposed to solve the backward compatibility? Probably it's my misconception about the solution, because I assumed that the users of the library are not using the sort package directly.

Also you can drop my PR, since it's only implements the function signature change. The harder part (what you need to take care of in this case) will be to discover where the caller functions are located to implement the defer recover logic.

alovak:
Hey! What I see from the code is that `panic` is invoked when we validate `field.Spec` or `MessageSpec`. It happens when we create a new field/message using the constructor. You can see here that, in the case of `Composite` field, it creates all [subfields](https://github.com/moov-io/iso8583/blob/master/field/spec.go#L101) and calls `SetSpec` for each of them. Note that some of these might also be composite fields, making the process recursive.

[...]

PR #275, opened 2023-09-07
Solve panic issues

alovak:
Closes #270 

With this PR we:
* introduce `Validate()` method for `MessageSpec` and `field.Spec` - so if you don't want constructor to panic, you can validate your spec before.
* in `StringsByHex` if sorted value is not a Hex, we compare them as regular strings (no `panic`)
* ignore `panic` in fuzzer, I hope later we can address it

### DIFF
--- a/message.go
+++ b/message.go
@@ -33,26 +33,18 @@ type Message struct {
 }
 
 func NewMessage(spec *MessageSpec) *Message {
-	// validate the spec
+	// Validate the spec
 	if err := spec.Validate(); err != nil {
 		panic(err) //nolint // as specs moslty static, we panic on spec validation errors
 	}
 
 	fields := spec.CreateMessageFields()
 
-	m := &Message{
+	return &Message{
 		fields:    fields,
 		spec:      spec,
 		fieldsMap: map[int]struct{}{},
 	}
-
-	// we validate the presence and type of the bitmap field
-	// in spec.Validate() so we can safely assume it exists
-	// and is of the correct type
-	m.bitmap, _ = m.fields[bitmapIdx].(*field.Bitmap)
-	m.fieldsMap[bitmapIdx] = struct{}{}
-
-	return m
 }
 
 // Deprecated. Use Marshal intead.
@@ -61,6 +53,17 @@ func (m *Message) SetData(data interface{}) error {
 }
 
 func (m *Message) Bitmap() *field.Bitmap {
+	if m.bitmap != nil {
+		return m.bitmap
+	}
+
+	// We validate the presence and type of the bitmap field in
+	// spec.Validate() when we create the message so we can safely assume
+	// it exists and is of the correct type
+	m.bitmap, _ = m.fields[bitmapIdx].(*field.Bitmap)
+	m.bitmap.Reset()
+	m.fieldsMap[bitmapIdx] = struct{}{}
+
 	return m.bitmap
 }
 
--- a/message.go
+++ b/message.go
@@ -33,13 +33,26 @@ type Message struct {
 }
 
 func NewMessage(spec *MessageSpec) *Message {
+	// validate the spec
+	if err := spec.Validate(); err != nil {
+		panic(err) //nolint // as specs moslty static, we panic on spec validation errors
+	}
+
 	fields := spec.CreateMessageFields()
 
-	return &Message{
+	m := &Message{
 		fields:    fields,
 		spec:      spec,
 		fieldsMap: map[int]struct{}{},
 	}
+
+	// we validate the presence and type of the bitmap field
+	// in spec.Validate() so we can safely assume it exists
+	// and is of the correct type
+	m.bitmap, _ = m.fields[bitmapIdx].(*field.Bitmap)
+	m.fieldsMap[bitmapIdx] = struct{}{}
+
+	return m
 }
 
 // Deprecated. Use Marshal intead.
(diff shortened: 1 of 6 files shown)


PumpkinSeed:
What do you think about creating a `NewMessage` version with error return? If someone wants to handle it on that way. (Speaking about my own case mostly) 

For exmaple:

```
func NewMessage(spec *MessageSpec) *Message {
	message, err := newMessage(spec)
	if err != nil {
		panic(err) //nolint // as specs moslty static, we panic on spec validation errors
	}
	return message
}

// NOTE: I don't know what would be a good name for that
// NOTE2: There should be an explanation about backward compatibility
func NewMessageWithError(spec *MessageSpec) (*Message, error) {
	return newMessage(spec)
}

func newMessage(spec *MessageSpec) (*Message, error) {
	// Validate the spec
	if err := spec.Validate(); err != nil {
		return nil, err
	}
	.
	.
	...
}
```

Let me know what you think about it.

PumpkinSeed:
Also, it closes #269, not #270. :)

alovak:
When you define spec, it will panic if you have a composite field with invalid spec. So, `panic` happens before `NewMessage`.

Panicking in Go is generally reserved for situations where it's inappropriate or impossible for the program to continue. In our case, if the spec is not valid, we can't proceed. I believe using `panic` is acceptable.

wadearnold (approved):

message.go
```diff
@@ -52,15 +57,10 @@ func (m *Message) Bitmap() *field.Bitmap {
...
-	if m.bitmap, ok = bitmap.(*field.Bitmap); !ok {
-		panic(fmt.Sprintf("bitmap field is %T not of type *field.Bitmap", m.fields[bitmapIdx]))
-	}
-
+	// We validate the presence and type of the bitmap field in
+	// spec.Validate() when we create the message so we can safely assume
+	// it exists and is of the correct type
+	m.bitmap, _ = m.fields[bitmapIdx].(*field.Bitmap)
 	m.bitmap.Reset()
 	m.fieldsMap[bitmapIdx] = struct{}{}
```
adamdecaf, line 65: Do you need to check if `m.bitmap` is non-nil before using it on L64?

adamdecaf (approved):

message.go
alovak: When we validate the spec we check that it defines bitmap, then NewMessage creates all fields defined in the spec, including bitmap. So we don't have to check it here, but it's a good catch!

PumpkinSeed:
> When you define spec, it will panic if you have a composite field with invalid spec. So, `panic` happens before `NewMessage`.
> 
> Panicking in Go is generally reserved for situations where it's inappropriate or impossible for the program to continue. In our case, if the spec is not valid, we can't proceed. I believe using `panic` is acceptable.

By default I agree on that usage of panic and I'm not trying to push my point of view, but I would like to share some argument to add an exported function with error return. (Note, I don't want to remove the original behavior, I'm just suggesting a minor extension)

First I would like to highlight the usage of the library. By following the [original documentation](https://github.com/moov-io/iso8583#build-and-pack-the-message) located in the README, we call the `message := NewMessage(spec)` with a specification defined as JSON and marshalled into the struct representation. We do that every time when we build a new transaction request and also we do that when an incoming message arrives. We do `Unpack` and `Pack`.

1. In idiomatic Go library users didn't expect that the library itself breaks the codes execution. That supposed to be the job of the programs control flow and it the writer of the program should be able to take control when the program break it's execution.
2. In our use-case we are talking about multiple unreliable TCP connection which handled by multiple concurrent workers in-memory queues for channels. (Probably it's our own failure to write something in that way) In this case one service handle multiple targets of iso8583 compliant backends. So getting a panic in such an unreliable environment puts extra job on the application, however if the error getting returned we could have handle it easily.
3. I know that the bad spec is not something what we need to execute anymore, but we have multiple different specs (because multiple payment processors has their own and we have at least 5 different one). That would be nice if in case of one bad specification it wouldn't break the entire application and let the scheduler decide whether it propagate the changes to the cluster based on the error log rate.

Based on these arguments I hope you can reconsider the suggestion above.

adamdecaf:
I want to understand what you're asking us to reconsider. Are you asking that methods like `NewMessage` and `SetSpec` return `error` instead of panic?

PumpkinSeed:
@adamdecaf 
I had this [comment](https://github.com/moov-io/iso8583/pull/275#issuecomment-1710043903) on this pull request.

PumpkinSeed:
@adamdecaf
Sorry I just answered quickly. By following the previously linked comment, I would like you to reconsider that you add an exported function which implements the same as `NewMessage`, but it returns with error instead of panic. I'm not the best in naming so I mentioned `NewMessageWithError`, but obviously it's not the best. Based on the reasons and arguments from above, this makes the user of the library to decide whether the program can handle a panic in case of any issue or they want to handle it themselves by using this new exported function. 

The main issue with the NewMessage that it has a lots of unexported part. So I can't write a function which is not relying on panic.

adamdecaf:
I don't think supporting backwards compatibility of panics is something we want to support. Causing a `panic` is very expensive and dangerous to systems. Callers can still panic if they wish, but our libraries always try to avoid that.

PumpkinSeed:
> I don't think supporting backwards compatibility of panics is something we want to support. Causing a `panic` is very expensive and dangerous to systems. Callers can still panic if they wish, but our libraries always try to avoid that.

I don't want to pick a side in this question, I just offered a possible solution where the caller of the library can decide between the backward compatible way or using another function with error return.

Otherwise I wrote a wrapper solution to avoid panic for our use-case. So I'm also fine with the current implementation as well.

PR #284, opened 2023-09-21
Make the package safe for concurrent access

alovak:
With this PR, we've enhanced the concurrency safety of the methods in `iso8583.Message` and `field.Composite`.

## Rationale

I encountered a data race when two goroutines executed `message.GetString(11)`. This happened because the GetString method did more than just returning a value — it also updated an internal map. This concurrent access led to the race condition.

## Assumptions on Concurrency:

* Individual field access (like `field.String`, `field.Numeric`) is not thread-safe. Concurrency safety is ensured at the `Message` and `Composite` levels. If external concurrent access to individual fields is necessary, users should manage it with appropriate synchronization mechanisms, such as a mutex.
* Concurrent access is expected to be infrequent. Given this, the overhead introduced by using a mutex within Message and Composite is likely to be minimal.

## Concerns

A potential solution to the data race I encountered would have been to simply eliminate the unnecessary map modification, thus making `GetXXX()` methods pure getters (which is also done in the PR). However, because the message is referenced via a pointer and has a combination of getters, setters, `Marshal`, and `Unmarshal` methods, it doesn't feel entirely safe to assume that access to these will always be sequential. So, enhancing concurrency safety seems like a guarded measure.

### DIFF
--- a/field/composite.go
+++ b/field/composite.go
@@ -359,8 +359,6 @@ func (f *Composite) Bitmap() *Bitmap {
 }
 
 func (f *Composite) bitmap() *Bitmap {
-	// TODO: protect against concurrent access
-
 	if f.cachedBitmap != nil {
 		return f.cachedBitmap
 	}
--- a/field/composite.go
+++ b/field/composite.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"regexp"
 	"strconv"
+	"sync"
 
 	"github.com/moov-io/iso8583/encoding"
 	"github.com/moov-io/iso8583/prefix"
@@ -63,10 +64,14 @@ var _ json.Unmarshaler = (*Composite)(nil)
 // For the sake of determinism, packing of subfields is executed in order of Tag
 // (using Spec.Tag.Sort) regardless of the value of Spec.Tag.Length.
 type Composite struct {
-	spec   *Spec
-	bitmap *Bitmap
+	spec         *Spec
+	cachedBitmap *Bitmap
 
 	orderedSpecFieldTags []string
+
+	// mu is used to synchronize access to the subfields and
+	// setSubfields maps when the composite is used concurrently
+	mu sync.Mutex
 
 	// stores all fields according to the spec
 	subfields map[string]Field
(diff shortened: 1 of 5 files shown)


field/composite.go
```diff
@@ -308,14 +343,26 @@ func (f *Composite) SetBytes(data []byte) error {
...
-	if f.bitmap != nil {
-		return f.bitmap
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	return f.bitmap()
+}
+
+func (f *Composite) bitmap() *Bitmap {
+	// TODO: protect against concurrent access
```
bpross, line 362: Did you miss this instance? I see that its used in other unexported methods.

field/composite_test.go
```diff
@@ -1892,3 +1894,60 @@ func TestComposite_getFieldIndexOrTag(t *testing.T) {
...
+		// packing and marshaling
+		data := &TLVTestData{
+			F9A:   NewHexValue("210720"),
+			F9F02: NewHexValue("000000000501"),
+		}
+
+		composite := NewComposite(tlvTestSpec)
+
+		wg := sync.WaitGroup{}
+		wg.Add(5)
```
bpross, line 1909: Nit: Do this inside of the go routine before the defer and change it to `wg.Add(1)` that way the for loop is the arbiter of modifying the waitgroup.

message.go
```diff
@@ -241,15 +302,19 @@ func (m *Message) unpack(src []byte) error {
 	return nil
 }
 
+// TODO: protect against concurrent access
```
bpross, line 305: Nit: you can remove this comment, since the mutex is being used.

message_test.go
```diff
@@ -90,6 +91,27 @@ func TestMessage(t *testing.T) {
...
 
+	// this test most probably will fail in regular mode,
+	// and should fail when is run with -race flag
+	t.Run("No data race when accessing fields concurrently", func(t *testing.T) {
+		message := NewMessage(spec)
+
+		var wg sync.WaitGroup
+
+		for i := 0; i < 1000; i++ {
+			wg.Add(1)
```
bpross, line 102: 👍 This matches what I was suggesting above.

bpross (approved):

wadearnold (approved):

Issue #278, opened 2023-09-08
Deprecate `SetData` method


alovak:
`SetData` is there to support compatibility with our first version of Marshal/Unmarshal implementation. In this [PR](https://github.com/moov-io/iso8583/pull/157) I described why using `Data` and `SetData` was not good. In short, it has some side effects such as linking objects using pointers.

The following should be done:
* get rid of the `data` struct field in all fields of the `field` package (`String`, `Numeric`, etc.)
* move code from `SetData` to `Marshal`
* deprecate `SetData` method

This change should not break any integrations unless they still use the version with the `Data()` method that was deprecated and deleted.

alovak:
closed in #288

PR #288, opened 2023-10-02
Deprecate SetData method

mfdeveloper508:
https://github.com/moov-io/iso8583/issues/278

### DIFF
--- a/field/binary.go
+++ b/field/binary.go
@@ -114,6 +114,11 @@ func (f *Binary) Unpack(data []byte) (int, error) {
 	return read + prefBytes, nil
 }
 
+// Deprecated. Use Marshal instead
+func (f *Binary) SetData(data interface{}) error {
+	return f.Marshal(data)
+}
+
 func (f *Binary) Unmarshal(v interface{}) error {
 	if v == nil {
 		return nil
@@ -129,12 +134,12 @@ func (f *Binary) Unmarshal(v interface{}) error {
 	return nil
 }
 
-func (f *Binary) SetData(data interface{}) error {
-	if data == nil {
+func (f *Binary) Marshal(v interface{}) error {
+	if v == nil {
 		return nil
 	}
 
-	bin, ok := data.(*Binary)
+	bin, ok := v.(*Binary)
 	if !ok {
 		return errors.New("data does not match required *Binary type")
 	}
@@ -144,10 +149,6 @@ func (f *Binary) SetData(data interface{}) error {
 		f.value = bin.value
 	}
 	return nil
-}
-
-func (f *Binary) Marshal(data interface{}) error {
-	return f.SetData(data)
 }
 
 func (f *Binary) MarshalJSON() ([]byte, error) {
--- a/field/bitmap.go
+++ b/field/bitmap.go
@@ -114,6 +114,11 @@ func (f *Bitmap) Unpack(data []byte) (int, error) {
 	return read, nil
 }
 
+// Deprecated. Use Marshal instead
+func (f *Bitmap) SetData(data interface{}) error {
+	return f.Marshal(data)
+}
+
 func (f *Bitmap) Unmarshal(v interface{}) error {
 	if v == nil {
 		return nil
@@ -129,22 +134,18 @@ func (f *Bitmap) Unmarshal(v interface{}) error {
 	return nil
 }
 
-func (f *Bitmap) SetData(data interface{}) error {
-	if data == nil {
+func (f *Bitmap) Marshal(v interface{}) error {
+	if v == nil {
 		return nil
 	}
 
-	bmap, ok := data.(*Bitmap)
+	bmap, ok := v.(*Bitmap)
 	if !ok {
 		return fmt.Errorf("data does not match required *Bitmap type")
 	}
 
 	f.data = bmap.data
 	return nil
-}
-
-func (f *Bitmap) Marshal(data interface{}) error {
-	return f.SetData(data)
 }
 
 // Reset resets the bitmap to its initial state because of how message works,
--- a/field/field.go
+++ b/field/field.go
@@ -23,7 +23,7 @@ type Field interface {
 	// Bytes returns binary representation of the field Value
 	Bytes() ([]byte, error)
 
-	// Deprecated. Use Marshal intead.
+	// Deprecated. Use Marshal instead.
 	SetData(data interface{}) error
 
 	// Unmarshal sets field Value into provided v. If v is nil or not
--- a/field/hex.go
+++ b/field/hex.go
@@ -126,6 +126,11 @@ func (f *Hex) Unpack(data []byte) (int, error) {
 	return read + prefBytes, nil
 }
 
+// Deprecated. Use Marshal instead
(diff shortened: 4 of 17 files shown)


field/binary.go
```diff
@@ -129,12 +134,12 @@ func (f *Binary) Unmarshal(v interface{}) error {
...
 
-func (f *Binary) SetData(data interface{}) error {
-	if data == nil {
+func (f *Binary) Marshal(v interface{}) error {
+	if v == nil {
 		return nil
 	}
 
-	bin, ok := data.(*Binary)
+	bin, ok := v.(*Binary)
```
alovak, line 142: @mfdeveloper508 here and in the rest of the fields, we don't need `f.data` anymore. Before, it was used in `SetData` and was a pointer that was used when we set and get "data". With Marhsal and Unmarshl - we don't need to store the pointer to the field type as we use field value directly.

```go
        f.data = bin // <- the field can be removed
	if bin.value != nil {
		f.value = bin.value
	}
```

field/bitmap.go
```diff
@@ -129,12 +134,12 @@ func (f *Bitmap) Unmarshal(v interface{}) error {
...
 
-func (f *Bitmap) SetData(data interface{}) error {
-	if data == nil {
+func (f *Bitmap) Marshal(v interface{}) error {
+	if v == nil {
 		return nil
 	}
 
-	bmap, ok := data.(*Bitmap)
+	bmap, ok := v.(*Bitmap)
```
alovak, line 142: In this field, we need `data` as it's basically `value` - it stores the value of the field. Please, don't be confused with my previous statement that we should remove `data` from all fields. When `data *String` (or pointer to other field type - then we should remove it), but here - `data` has the same purpose as `value` in other fields. It just has different name.

field/binary.go
mfdeveloper508: https://github.com/moov-io/iso8583/pull/289
right fixed in next pr

field/field.go
```diff
@@ -23,7 +23,7 @@ type Field interface {
 	// Bytes returns binary representation of the field Value
 	Bytes() ([]byte, error)
 
-	// Deprecated. Use Marshal intead.
+	// Deprecated. Use Marshal instead.
```
alovak, line 26: Good catch!

field/hex.go
```diff
@@ -141,12 +146,12 @@ func (f *Hex) Unmarshal(v interface{}) error {
...
 
-func (f *Hex) SetData(data interface{}) error {
-	if data == nil {
+func (f *Hex) Marshal(v interface{}) error {
+	if v == nil {
 		return nil
 	}
 
-	str, ok := data.(*Hex)
+	str, ok := v.(*Hex)
```
alovak, line 154: below, we can remove `f.data`

field/numeric.go
```diff
@@ -142,12 +147,12 @@ func (f *Numeric) Unmarshal(v interface{}) error {
...
 
-func (f *Numeric) SetData(data interface{}) error {
-	if data == nil {
+func (f *Numeric) Marshal(v interface{}) error {
+	if v == nil {
 		return nil
 	}
 
-	num, ok := data.(*Numeric)
+	num, ok := v.(*Numeric)
```
alovak, line 155: here we can remove `data` as well

field/numeric_test.go
```diff
@@ -37,14 +37,14 @@ func TestNumericField(t *testing.T) {
...
-	numeric.SetData(NewNumericValue(9876))
+	numeric.Marshal(NewNumericValue(9876))
 	packed, err = numeric.Pack()
 	require.NoError(t, err)
 	require.Equal(t, "      9876", string(packed))
 
 	numeric = NewNumeric(spec)
 	data := NewNumericValue(0)
-	numeric.SetData(data)
+	numeric.Marshal(data)
```
alovak, line 47: This test will fail after we remove `data` from the `Numeric` field. The behavior of `SetData` is slightly different from `Marshal`. Just test that `Marshal` sets value and `Unmarshal` sets value back to the arg.

field/string.go
```diff
@@ -128,12 +133,12 @@ func (f *String) Unmarshal(v interface{}) error {
...
 
-func (f *String) SetData(data interface{}) error {
-	if data == nil {
+func (f *String) Marshal(v interface{}) error {
+	if v == nil {
 		return nil
 	}
 
-	str, ok := data.(*String)
+	str, ok := v.(*String)
```
alovak, line 141: `f.data` can be removed from the `String` field

field/string_test.go
```diff
@@ -37,22 +37,22 @@ func TestStringField(t *testing.T) {
...
 	require.Equal(t, "     hello", string(packed))
 
 	str = NewString(spec)
 	data := NewStringValue("")
-	str.SetData(data)
+	str.Marshal(data)
 	length, err = str.Unpack([]byte("     olleh"))
 	require.NoError(t, err)
 	require.Equal(t, 10, length)
 	require.Equal(t, "olleh", data.Value())
```
alovak, line 51: This test will also fail as if we remove `data` it will no longer be used in `Unpack` (internally `SetBytes`). Just test `Marshal` and `Unmarshal`.

field/track1.go
```diff
@@ -132,12 +137,12 @@ func (f *Track1) Unmarshal(v interface{}) error {
...
 
-func (f *Track1) SetData(data interface{}) error {
-	if data == nil {
+func (f *Track1) Marshal(v interface{}) error {
+	if v == nil {
 		return nil
 	}
 
-	track, ok := data.(*Track1)
+	track, ok := v.(*Track1)
```
alovak, line 145: `f.data` can be removed from the Track1

field/track2.go
```diff
@@ -128,12 +133,12 @@ func (f *Track2) Unmarshal(v interface{}) error {
...
 
-func (f *Track2) SetData(data interface{}) error {
-	if data == nil {
+func (f *Track2) Marshal(v interface{}) error {
+	if v == nil {
 		return nil
 	}
 
-	track, ok := data.(*Track2)
+	track, ok := v.(*Track2)
```
alovak, line 141: `f.data` can be removed from `Track2`

field/track3.go
```diff
@@ -125,12 +130,12 @@ func (f *Track3) Unmarshal(v interface{}) error {
...
 
-func (f *Track3) SetData(data interface{}) error {
-	if data == nil {
+func (f *Track3) Marshal(v interface{}) error {
+	if v == nil {
 		return nil
 	}
 
-	track, ok := data.(*Track3)
+	track, ok := v.(*Track3)
```
alovak, line 138: here we can also remove `f.data`.

alovak:
@mfdeveloper508 thanks for breaking down your PR into smaller ones. For this PR the summary of my review is:
1. we should remove data field from all fields such as Numeric, String, etc. Please note, that in Binary field, it should stay or be renamed into value as it’s named in all fields.
2. we should update some tests as after we remove data - I believe 2 tests will fail. Just simplify them to keep tests only for Marshal and Unmarshal

Issue #282, opened 2023-09-11
Update `Unmarshal` method for the message and the rest of the fields


alovak:
It should be possible to use native types in data structures like this when we extract data from the message into the struct using `Unmarshal`:

```go
// define a struct for the authorization request
type AuthorizationResponse struct {
	MTI                  string `index:"0"`
	STAN                 string `index:"11"`
	// ...
}

// usage
request := &AuthorizationResponse{}

message := iso8583.NewMessage(Spec)
err := message.Unpack(someBinaryData)
// handle error
err = message.Unmarshal(request)
// handle error

// request.MTI => "0110"
// request.STAN => "000123"
```

It should include the following:
* `Message`
* `field.String` -> supported target types string, int (+pointers)
* `field.Numeric` -> supported target types string, int (+pointers)
* `field.Binary` -> supported target types are `[]byte`, `string`, `*string`. If target is`string`, we should convert the value into hex
* `field.Hex`-> same as `field.Binary`

the reference implementation is here and it includes update of the `message.Unmarshal` ([src](https://github.com/moov-io/iso8583/pull/273/files#diff-2ce052ca28a77c3b54be26bfc568dddef91bf239253d0ad537609af1731032d4R437)) and `field.String`/`Unmarshal` ([src](https://github.com/moov-io/iso8583/pull/273/files#diff-79ac174d65bca7ada39659ad4ae9fb2ae82b4532ca22f9392010bad81638708cR113)):

alovak:
closed by #291

Issue #279, opened 2023-09-08
Update `Marshal` method for the fields to accept Go native types


alovak:
We want to support Go's native types such as `string`, `int`, etc. when we set data to the field using `Marshal` method

So, the idea is to implement the following:

```
f := field.NewString(someSpec)

err := f.Marshal("hello")
f.Value() // "hello"

err := f.Marshal(5)
f.Value() // "5"

str := "hello"
err := f.Marshal(&str)
f.Value() // "hello"

i := 5
err := f.Marshal(&i)
f.Value() // "5"
```

We want to be able to use:
* `int`, `*int`, `string`, `*string` for `field.Numeric`, `field.String`
* `[]byte`, `string` for `field.Binary`, `field.Hex` (`string` should be hex and if we fail to decode it, return error)

Reference implementation is in this PR: https://github.com/moov-io/iso8583/pull/273/files

alovak:
closed by #289

PR #289, opened 2023-10-02
Updating unimarshaller

mfdeveloper508:
https://github.com/moov-io/iso8583/issues/282
https://github.com/moov-io/iso8583/issues/279

### DIFF
--- a/field/binary.go
+++ b/field/binary.go
@@ -137,8 +137,7 @@ func (f *Binary) Unmarshal(v interface{}) error {
 			return fmt.Errorf("unsupported reflect.Value type: %s", val.Kind())
 		}
 	case *string:
-		str := hex.EncodeToString(f.value)
-		*val = str
+		*val = hex.EncodeToString(f.value)
 	case *[]byte:
 		*val = f.value
 	case *Binary:
--- a/field/binary.go
+++ b/field/binary.go
@@ -1,9 +1,10 @@
 package field
 
 import (
+	"encoding/hex"
 	"encoding/json"
-	"errors"
 	"fmt"
+	"reflect"
 
 	"github.com/moov-io/iso8583/encoding"
 	"github.com/moov-io/iso8583/utils"
@@ -16,7 +17,6 @@ var _ json.Unmarshaler = (*Binary)(nil)
 type Binary struct {
 	value []byte
 	spec  *Spec
-	data  *Binary
 }
 
 func NewBinary(spec *Spec) *Binary {
@@ -41,9 +41,6 @@ func (f *Binary) SetSpec(spec *Spec) {
 
 func (f *Binary) SetBytes(b []byte) error {
 	f.value = b
-	if f.data != nil {
-		*(f.data) = *f
-	}
 	return nil
 }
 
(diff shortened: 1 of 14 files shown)


field/binary.go
```diff
@@ -120,34 +117,82 @@ func (f *Binary) SetData(data interface{}) error {
 }
 
 func (f *Binary) Unmarshal(v interface{}) error {
-	if v == nil {
-		return nil
+	switch val := v.(type) {
+	case reflect.Value:
+		switch val.Kind() { //nolint:exhaustive
```
alovak, line 122:
```suggestion
	case reflect.Value:
		if !val.CanSet() {
			return fmt.Errorf("cannot set reflect.Value of type %s", val.Kind())
		}

		switch val.Kind() { //nolint:exhaustive
```

field/binary.go
```diff
@@ -120,34 +117,82 @@ func (f *Binary) SetData(data interface{}) error {
...
+		case reflect.String:
+			if !val.CanSet() {
+				return fmt.Errorf("reflect.Value of the data can not be change")
+			}
```
alovak, line 126:
```suggestion
```

field/binary.go
```diff
@@ -120,34 +117,82 @@ func (f *Binary) SetData(data interface{}) error {
...
+
+			str := hex.EncodeToString(f.value)
+			val.SetString(str)
+		case reflect.Slice:
+			if !val.CanSet() {
+				return fmt.Errorf("reflect.Value of the data can not be change")
+			}
+
```
alovak, line 134:
```suggestion
```

field/binary.go
```diff
@@ -120,34 +117,82 @@ func (f *Binary) SetData(data interface{}) error {
...
+			val.SetBytes(f.value)
+		default:
+			return fmt.Errorf("data does not match required reflect.Value type")
+		}
+	case *string:
+		str := hex.EncodeToString(f.value)
+		*val = str
```
alovak, line 141:
```suggestion
		*val = hex.EncodeToString(f.value)
```

field/binary.go
```diff
@@ -120,34 +117,82 @@ func (f *Binary) SetData(data interface{}) error {
...
+			str := hex.EncodeToString(f.value)
+			val.SetString(str)
+		case reflect.Slice:
+			if !val.CanSet() {
+				return fmt.Errorf("reflect.Value of the data can not be change")
+			}
+
+			val.SetBytes(f.value)
+		default:
+			return fmt.Errorf("data does not match required reflect.Value type")
```
alovak, line 137:
```suggestion
			return fmt.Errorf("unsupported reflect.Value type: %s", val.Kind())
```

field/binary.go
```diff
@@ -120,34 +117,82 @@ func (f *Binary) SetData(data interface{}) error {
...
+		}
+	case *string:
+		str := hex.EncodeToString(f.value)
+		*val = str
+	case *[]byte:
+		*val = f.value
+	case *Binary:
+		val.value = f.value
+	default:
+		return fmt.Errorf("data does not match required *Binary or (*string, *[]byte) type")
```
alovak, line 147:
```suggestion
		return fmt.Errorf("unsupported type: expected *Binary, *string, *[]byte, or reflect.Value, got %T", v)
```

field/binary.go
```diff
@@ -120,34 +117,82 @@ func (f *Binary) SetData(data interface{}) error {
...
+		}
+
+		buf, err := hex.DecodeString(*v)
+		if err != nil {
+			return fmt.Errorf("failed to convert string to byte: %w", err)
+		}
+
+		f.value = buf
+	case []byte:
+		f.SetBytes(v)
```
alovak, line 185: It can also be `nil`.

field/binary.go
```diff
@@ -120,34 +117,82 @@ func (f *Binary) SetData(data interface{}) error {
...
+	case *[]byte:
+		if v == nil {
+			f.value = nil
+			return nil
+		}
+		f.SetBytes(*v)
+	default:
+		return fmt.Errorf("data does not match required *Binary or (string, *string, []byte, *[]byte) type")
```
alovak, line 196: @mfdeveloper508 I suggest us to rework the method a little bit. First, we always set f.value to nil if `v` has zero value and return. So we can do this only once and the rest of the code will not have `if` statements. It seems the following code should work:


```go
func (f *Binary) Marshal(v interface{}) error {
	if reflect.ValueOf(v).IsZero() {
		f.value = nil
		return nil
	}

	switch val := v.(type) {
	case *Binary:
		f.value = val.value
	case string:
		buf, err := hex.DecodeString(val)
		if err != nil {
			return utils.NewSafeError(fmt.Errorf("converting string (as hex) to byte: %w", err), "failed to convert string (as hex) to byte") 
		}
		f.value = buf
	case *string:
		buf, err := hex.DecodeString(*val)
		if err != nil {
			return utils.NewSafeError(fmt.Errorf("converting string (as hex) to byte: %w", err), "failed to convert string (as hex) to byte") 
		}
		f.value = buf
	case []byte:
		f.SetBytes(val)
	case *[]byte:
		f.SetBytes(*val)
	default:
		return fmt.Errorf("data does not match required *Binary or (string, *string, []byte, *[]byte) type")
	}

	return nil
}
```

field/binary_test.go
```diff
@@ -140,3 +141,76 @@ func TestBinaryNil(t *testing.T) {
...
+	err = str.Unmarshal(val5)
+	require.NoError(t, err)
+	require.Equal(t, testValue, val5.Bytes())
+
+	val6 := reflect.ValueOf(val2)
+	err = str.Unmarshal(val6)
+	require.Error(t, err)
+	require.Equal(t, "cannot set reflect.Value of type string", err.Error())
+
+	val7 := reflect.ValueOf(&val2)
```
alovak, line 179: I would suggest to name variables a little bit differently. In this line, it's not clear what is the type of `val2`.

What about the following changes:

```
testValue := []byte{0x12, 0x34, 0x56} -> expectedBytes
val1 := &Binary{} -> binary or binaryField
var val2 string -> str or s
val3 []byte -> bytes
```

If you agree, we can do this for all tests.

mfdeveloper508: agree

alovak (approved):
Great job!

PR #291, opened 2023-10-05
Update message unimarshaller functions

mfdeveloper508:
https://github.com/moov-io/iso8583/issues/282

### DIFF
--- a/field/composite.go
+++ b/field/composite.go
@@ -206,14 +206,10 @@ func (f *Composite) Unmarshal(v interface{}) error {
 				return fmt.Errorf("failed to get data from field %s: %w", indexOrTag, err)
 			}
 		default: // Native types
-			// using new reflect value that has pointer instead of input data field
-			dummy := reflect.New(dataField.Type()).Elem()
-			err = messageField.Unmarshal(dummy)
+			err = messageField.Unmarshal(dataField)
 			if err != nil {
 				return fmt.Errorf("failed to get data from field %s: %w", indexOrTag, err)
 			}
-
-			dataField.Set(dummy)
 		}
 	}
 
--- a/field/composite.go
+++ b/field/composite.go
@@ -195,13 +195,23 @@ func (f *Composite) Unmarshal(v interface{}) error {
 		}
 
 		dataField := dataStruct.Field(i)
-		if dataField.IsNil() {
-			dataField.Set(reflect.New(dataField.Type().Elem()))
-		}
+		switch dataField.Kind() { //nolint:exhaustive
+		case reflect.Chan, reflect.Func, reflect.Map, reflect.Pointer, reflect.UnsafePointer, reflect.Interface, reflect.Slice:
+			if dataField.IsNil() {
+				dataField.Set(reflect.New(dataField.Type().Elem()))
+			}
 
-		err = messageField.Unmarshal(dataField.Interface())
-		if err != nil {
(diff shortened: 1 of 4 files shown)


field/composite.go
```diff
@@ -195,13 +195,25 @@ func (f *Composite) Unmarshal(v interface{}) error {
...
-		err = messageField.Unmarshal(dataField.Interface())
-		if err != nil {
-			return fmt.Errorf("failed to get data from field %s: %w", indexOrTag, err)
+			err = messageField.Unmarshal(dataField.Interface())
+			if err != nil {
+				return fmt.Errorf("failed to get data from field %s: %w", indexOrTag, err)
+			}
+		default: // Native types
+			// using new reflect value that has pointer instead of input data field
+			dummy := reflect.New(dataField.Type()).Elem()
```
alovak, line 210: I'm not sure I understand the need of `dummy` here.

message.go
```diff
@@ -502,13 +502,25 @@ func (m *Message) Unmarshal(v interface{}) error {
...
-		err = messageField.Unmarshal(dataField.Interface())
-		if err != nil {
-			return fmt.Errorf("failed to get value from field %d: %w", fieldIndex, err)
+			err = messageField.Unmarshal(dataField.Interface())
+			if err != nil {
+				return fmt.Errorf("failed to get value from field %d: %w", fieldIndex, err)
+			}
+		default: // Native types
+			// using new reflect value that has pointer instead of input data field
+			dummy := reflect.New(dataField.Type()).Elem()
```
alovak, line 517: Same here

alovak:
It looks good! Would you please add table tests for Marshal/Unmarshal something like this this:https://github.com/moov-io/iso8583/blob/2f44c9f5c9bf72927313d35e4dfaa478711943fb/exp_test.go?

message.go
```diff
@@ -502,13 +502,25 @@ func (m *Message) Unmarshal(v interface{}) error {
...
+			err = messageField.Unmarshal(dummy)
+			if err != nil {
+				return fmt.Errorf("failed to get value from field %d: %w", fieldIndex, err)
+			}
+
+			dataField.Set(dummy)
```
alovak, line 523: If you remove `dummy` and keep code like this, all tests will work:

```suggestion
			err = messageField.Unmarshal(dataField)
			if err != nil {
				return fmt.Errorf("failed to get value from field %d: %w", fieldIndex, err)
			}
```

it's because in `dataField := dataStruct.Field(i)` `dataField` is `reflect.Value`, which we are using inside each field.

field/string.go
```diff
@@ -153,7 +153,7 @@ func (f *String) Unmarshal(v interface{}) error {
 }
 
 func (f *String) Marshal(v interface{}) error {
-	if v == nil || reflect.ValueOf(v).IsZero() {
+	if v == nil || (!reflect.ValueOf(v).CanInt() && reflect.ValueOf(v).IsZero()) {
```
alovak, line 156: is there a test that can explain this change? I think that without `keepzero` this can stay as is.

But when we add `keepzero` we should check if value is `0` (int/*int) for `field.String`, then we should assign it.

Also, it may be similar to some other types. Not sure.

alovak (approved):

field/string.go
mfdeveloper508: right 
the code is exception for having int value zero

Issue #281, opened 2023-09-08
Support new tag `iso8583` in addition to `index`


alovak:
Using tag `index` was not a good idea, typically, tags named as `json`, `xml`, `sql`. In the context of this package, it should be named `iso8583`. Also, when we add more attributes to the tag like `"3,keepzero"`,  the name `index` does not make any sense.

We should support the following: `Amount int `iso8583:"3,keepzero"` and keep `index` for compatibility.

POC for this is implemented here: https://github.com/moov-io/iso8583/pull/273/files

GalihFajar:
hi @alovak I'd like to work on this issue (I'm currently delving into the repository so can't promise much 🙇 ).

alovak:
@GalihFajar afaik @mfdeveloper508 is working on this issue. You can look at this issue: https://github.com/moov-io/iso8583/issues/292. It does not require a lot of change and it seems it's a good for the first commit.

GalihFajar:
Thanks @alovak, will look onto that one

alovak:
closed in #293

PR #293, opened 2023-10-06
support index tag

mfdeveloper508:
https://github.com/moov-io/iso8583/issues/281


message.go
```diff
@@ -481,74 +485,39 @@ func (m *Message) Unmarshal(v interface{}) error {
...
-		if _, set := m.fieldsMap[fieldIndex]; !set {
+		if _, set := m.fieldsMap[indexTag.Id]; !set {
 			continue
 		}
 
 		dataField := dataStruct.Field(i)
 		switch dataField.Kind() { //nolint:exhaustive
 		case reflect.Chan, reflect.Func, reflect.Map, reflect.Pointer, reflect.UnsafePointer, reflect.Interface, reflect.Slice:
-			if dataField.IsNil() {
+			if dataField.IsNil() || dataField.IsZero() {
```
alovak, line 507: I'm curious, why do we check for zero value here?

exp_test.go
```diff
@@ -0,0 +1,228 @@
+package iso8583_test
```
alovak, line 1: May I ask you to move the tests into `message_test` and to review if all test cases you may think about are covered by these tests?

exp_test.go
```diff
@@ -0,0 +1,228 @@
...
+				}{
+					MTI: "0110",
+				},
+				expectedPackedString: "011000000000000000000000000000000000",
+			},
+			{
+				name: "struct with *int type, no value and keepzero tag",
+				input: struct {
+					MTI                  string `index:"0"`
+					PrimaryAccountNumber *int   `index:"2,keepzero"`
```
alovak, line 147: same here. I think we should create a new **int** and set the field value to `0`.

message.go
```diff
@@ -481,74 +485,39 @@ func (m *Message) Unmarshal(v interface{}) error {
...
 		}
 
-		if _, set := m.fieldsMap[fieldIndex]; !set {
+		if _, set := m.fieldsMap[indexTag.Id]; !set {
 			continue
 		}
 
 		dataField := dataStruct.Field(i)
 		switch dataField.Kind() { //nolint:exhaustive
 		case reflect.Chan, reflect.Func, reflect.Map, reflect.Pointer, reflect.UnsafePointer, reflect.Interface, reflect.Slice:
```
alovak, line 506: This comment a bit late and is not about the scope of this ticket, but about the previous PR. Looking at this line I think we should be specific about what do we support:
```suggestion
		case reflect.Pointer, reflect.Interface, reflect.Slice:
```

the same should be for other types (Int, Int64, String)
and finally, in the `default` we should return the error.

What do you think? We can create a separate ticket for this change.

exp_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+				packed, err := message.Pack()
+				require.NoError(t, err)
+
+				require.Equal(t, tt.expectedPackedString, string(packed))
+			})
+		}
+	})
+
+	t.Run("unpack", func(t *testing.T) {
```
alovak, line 169: I would review `Unmarshal`code and write tests to cover all `if` and `case` blocks such as:
* test for setting []byte

and also test negative scenarios:
* using map
* use slice of strings
* etc.

message_test.go
```diff
@@ -1415,21 +1415,17 @@ func Test_getFieldIndex(t *testing.T) {
...
 
-		index, err := getFieldIndex(st.Type().Field(0))
-
-		require.NoError(t, err)
-		require.Equal(t, 1, index)
+		indexTag := field.NewIndexTag(st.Type().Field(0))
+		require.Equal(t, 1, indexTag.Id)
 	})
 
 	t.Run("returns index from field tag instead of field name when both match", func(t *testing.T) {
```
alovak, line 1422: as we now have `field/index_tag.go` I think we should move all related tests into `field/index_tag_test.go` and remove them from other places. In index tag test we can test different test cases as well using table tests.

PR #298, opened 2023-11-06
Fix encoding non-UTF8 values in String field

### LINKED ISSUE #297: String Field does not handle non-UTF8 values correctly
The String field (and potentially other fields) does not handle the encoding of non-UTF8 values correctly, leading to incorrect lengths.

For example, take the value `hüllo`, and encoding EBCDIC 1047.

In UTF-8 hexadecimal bytes, this is `0x68 0xC3 0xBC 0x6C 0x6C 0x6F`, where `ü` is two bytes `0xC3 0xBC`.
In EBCDIC 1047, this is `0x88 0xDC 0x93 0x93 0x96`, where `ü` is one byte `0xDC`.

In the String field, padding is done before the data is encoded, and assumes the length of the data in UTF8 will be the same as the length of the data in the encoded format.

This means that when we specify a fixed length of 10 with padding, we get `0x40 0x40 0x40 0x40 0x88 0xDC 0x93 0x93 0x96`, which is only length 9.
When we specify a variable LL length, we get `0xF0 0xF6 0x88 0xDC 0x93 0x93 0x96`, which encoded that it has a length of 6 `0xF0 0xF6`, yet the actual data only has a length of 5.

```golang
spec := &Spec{
	Length:      10,
	Description: "Field",
	Enc:         encoding.EBCDIC1047,
	Pref:        prefix.EBCDIC1047.Fixed,
	Pad:         padding.Left(' '),
}
str := NewString(spec)
str.SetValue("hüllo")

packed, _ := str.Pack()
fmt.Printf("%X\n", packed) // 4040404088DC939396

lengthedSpec := &Spec{
	Length:      10,
	Description: "Field",
	Enc:         encoding.EBCDIC1047,
	Pref:        prefix.EBCDIC1047.LL,
}
str = NewString(lengthedSpec)
str.SetValue("hüllo")
packed, _ = str.Pack()
fmt.Printf("%X\n", packed) // F0F688DC939396
```

cheukwing:
Fixes #297 

One outstanding issue is what behaviour do we expect from `Bytes()` and `SetBytes()`?
Currently, these return/take the UTF8 encoded bytes, regardless of the specified encoding, but it seems like it could be more appropriate to return/take bytes in the specified encoding.
I tried doing this, but this led to issues with other functions in the library which expected the current behaviour, and users of this library may expect the same.

### DIFF
--- a/field/string.go
+++ b/field/string.go
@@ -70,18 +70,28 @@ func (f *String) SetValue(v string) {
 }
 
 func (f *String) Pack() ([]byte, error) {
-	data := []byte(f.value)
+	utf8Data := []byte(f.value)
 
 	if f.spec.Pad != nil {
-		data = f.spec.Pad.Pad(data, f.spec.Length)
+		// The length of the encoded data may differ from the UTF-8 encoded length.
+		// Use the difference to ensure the correct padded length.
+		unpaddedPacked, err := f.spec.Enc.Encode(utf8Data)
+		if err != nil {
+			return nil, fmt.Errorf("failed to encode content: %w", err)
+		}
+		diff := len(utf8Data) - len(unpaddedPacked)
+		if diff < 0 {
+			diff = -diff
+		}
+		utf8Data = f.spec.Pad.Pad(utf8Data, diff+f.spec.Length)
 	}
 
-	packed, err := f.spec.Enc.Encode(data)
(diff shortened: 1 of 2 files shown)


cheukwing:
Closing in favour of #299

PR #299, opened 2023-11-08
use runes

alovak:


cheukwing:
Thank you, this is much simpler than my fix!
We should also fix `padding/right.go`, and add some unit tests to the padders to finish off this draft.

alovak:
@cheukwing do you have a capacity to work on this feature taking this PR as a base? You better know what it should do as I first time meet the need to use utf8 :)

cheukwing:
I will have some capacity in a couple of weeks, but I've also asked some of my team members if they could help out too :slightly_smiling_face:

alovak:
I don't see any good options to support utf8 for now

PR #301, opened 2023-11-29
Add new field for HexNumerics

madhr:
Add new field type HexNumeric that allows to store binary encoded fields as hex numeric fields.

It's related to this change:
https://github.com/moov-io/iso8583/pull/227
where it's now possible to treat binary encoded fields as Hex.
Since Hex is a string field, it's a bit inconvenient for fields that store numeric values like amounts (for instance DE55.9F02).
Introducing the HexNumeric allows to convert the binary encoded fields straight into a hex numeric field.

### DIFF
--- a/field/hex_numeric.go
+++ b/field/hex_numeric.go
@@ -1,0 +1,262 @@
+package field
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strconv"
+
+	"github.com/moov-io/iso8583/utils"
+)
+
+var _ Field = (*HexNumeric)(nil)
+var _ json.Marshaler = (*HexNumeric)(nil)
+var _ json.Unmarshaler = (*HexNumeric)(nil)
+
+// HexNumeric is a Hex field that should be represented numerically.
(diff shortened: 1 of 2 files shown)


alovak:
Here is how you can get the same results using BCD encoding:

```go
func TestHexNumeric(t *testing.T) {
	numeric := NewNumeric(&Spec{
		Length:      4,
		Description: "Field",
		Enc:         encoding.BCD,
		Pref:        prefix.BCD.Fixed,
	})

	numeric.SetValue(1234)

	packed, err := numeric.Pack()
	require.NoError(t, err)

	require.Equal(t, []byte{0x12, 0x34}, packed)

	numeric2 := NewNumeric(&Spec{
		Length:      4,
		Description: "Field",
		Enc:         encoding.BCD,
		Pref:        prefix.BCD.Fixed,
	})

	_, err = numeric2.Unpack(packed)
	require.NoError(t, err)

	require.Equal(t, int64(1234), numeric2.Value())
}
```

alovak:
@madhr does the snippet I shared above work for you?

madhr:
> @madhr does the snippet I shared above work for you?

yes thanks!

PR #304, opened 2024-01-16
fix : noneprefixer did not return correct fixed len

T-eli:
this pull request contains 2 fixes:
1. noneprefixer fixed was returning len(data) instead of fixed value
2. text edit: numeric field parses int64 not int

while i am here i have a question if the someone can answer:
 -> Aren't  Numeric Fields usually encoded in BCD?   

Nemric field pack method:  `data := []byte(strconv.FormatInt(f.value, 10))... `

should be  :  `data, _ := hex.DecodeString(strconv.FormatInt(f.value, 10))`

I know that you can set the Encoder in spec to BCD.  but I dont see the deference between using a String Field and Numeric Field.

### DIFF
--- a/prefix/none.go
+++ b/prefix/none.go
@@ -12,7 +12,7 @@ func (p *nonePrefixer) EncodeLength(int, int) ([]byte, error) {
 }
 
 func (p *nonePrefixer) DecodeLength(fixLen int, data []byte) (int, int, error) {
-	return len(data), 0, nil
+	return fixLen, 0, nil
 }
 
 func (p *nonePrefixer) Inspect() string {
--- a/prefix/none.go
+++ b/prefix/none.go
@@ -12,7 +12,7 @@ func (p *nonePrefixer) EncodeLength(int, int) ([]byte, error) {
 }
 
 func (p *nonePrefixer) DecodeLength(fixLen int, data []byte) (int, int, error) {
-	return fixLen, 0, nil
+	return len(data), 0, nil
 }
 
 func (p *nonePrefixer) Inspect() string {
--- a/field/numeric.go
+++ b/field/numeric.go
@@ -185,7 +185,7 @@ func (f *Numeric) Marshal(v any) error {
 		}
 		f.value = val
 	default:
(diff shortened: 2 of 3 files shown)


alovak:
Hey! Thanks for the PR and the fixes!




> while i am here i have a question if the someone can answer: -> Aren't Numeric Fields usually encoded in BCD?
> 
> Nemric field pack method: `data := []byte(strconv.FormatInt(f.value, 10))... `
> 
> should be : `data, _ := hex.DecodeString(strconv.FormatInt(f.value, 10))`
> 
> I know that you can set the Encoder in spec to BCD. but I dont see the deference between using a String Field and Numeric Field.


I saw different encodings used for the numeric fields. It depends on the card brand / processor.

The difference between `String` and `Numeric` is in the type of the underlying value `string` vs `int64`:

```go
str := field.NewStringValue("100")
str.Value() // "100"

i := field.NewNumericValue(100)
i.Value() // 100
```

There will be an error if you try to unpack a message with a non-numeric field value for a field specified as `field.Numeric`.

prefix/none.go
```diff
@@ -12,7 +12,7 @@ func (p *nonePrefixer) EncodeLength(int, int) ([]byte, error) {
 }
 
 func (p *nonePrefixer) DecodeLength(fixLen int, data []byte) (int, int, error) {
-	return len(data), 0, nil
+	return fixLen, 0, nil
```
alovak, line 15: `None` prefixer is used in cases when you want to work with the value of the field and when you don't know the length upfront. If you know the length upfront - just use `ASCII.Fixed`.

Let's revert this change.

alovak:
We should revert the change to the `None` prefixer before we can merge the fixes.

prefix/none.go
T-eli: done

T-eli:
hello @alovak , 

thank you for answering my question.

i was confused by the `none.fixed`  which i interpreted as no prefix + use fixed length

anyways i cherry picked to other changes. if you want to merge this PR

alovak (approved):
@T-eli thanks a lot for the fixes!

T-eli:
thanks @alovak.

PR #313, opened 2024-07-12
Return subfields in unpack error

meparle:
Adding the capability to return subfield IDs from a composite type field within an `UnpackError` when there is an error unpacking a subfield.

It is necessary to provide this further detail for some card schemes as part of the format error information.

Adding a field `FieldIDs` to `UnpackError` to return the top level field ID plus an ID for each of however many levels of nested subfields error during unpacking.

The field `FieldID` could be deprecated once `FieldIDs` is present and sufficient notice given. Happy to add a deprecation notice if you would like.


field/composite.go
```diff
@@ -515,7 +516,28 @@ func (f *Composite) packByTag() ([]byte, error) {
...
+		var unpackErr *mooverrors.UnpackError
+		if errors.As(err, &unpackErr) {
+			subfields = unpackErr.FieldIDs
+		}
+		return offset, &mooverrors.UnpackError{
+			Err:        err,
+			FieldID:    tagID,
+			FieldIDs:   append([]string{tagID}, subfields...),
+			RawMessage: src,
+		}
```
alovak, line 535:
```suggestion
		return offset, &mooverrors.UnpackError{
			Err:        err,
			FieldID:    tagID,
			RawMessage: src,
		}
```

please, read the main comment that will explain this suggestion

message.go
```diff
@@ -238,9 +239,15 @@ func (m *Message) Unpack(src []byte) error {
...
+		var unpackErr *iso8583errors.UnpackError
+		if errors.As(err, &unpackErr) {
+			subfields = unpackErr.FieldIDs
+		}
+		return &iso8583errors.UnpackError{
 			Err:        err,
 			FieldID:    fieldID,
+			FieldIDs:   append([]string{fieldID}, subfields...),
 			RawMessage: src,
 		}
```
alovak, line 252:
```suggestion
		return &iso8583errors.UnpackError{
			Err:        err,
			FieldID:    fieldID,
			RawMessage: src,
		}
```

please, read the main message that will explain the suggestion

alovak (commented):
(main message)

what I suggest is to keep error wrapping you added, but instead of collecting the IDs in the wrapper methods, create a helper, that will iterate over the errors and extract all field IDs like this:

```
		fieldIDs := []string{}
		var unpackError *mooverrors.UnpackError
		for {
			if errors.As(err, &unpackError) {
				fieldIDs = append(fieldIDs, unpackError.FieldID)
				err = unpackError.Unwrap()
			} else {
				break
			}
		}

		fmt.Println(fieldIDs) // [3 2] - the same we expect in the tests
```

With such an approach, we keep UnpackError as it is, add error wrapping into a composite field and maybe a  helper method without breaking changes.

What do you think?

errors/errors.go
```diff
@@ -0,0 +1,32 @@
+package errors
```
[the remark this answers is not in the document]
meparle: Noting that we discussed this and agreed it makes sense to make this breaking change. The alternative would be to introduce another error type within field.

alovak (approved):

PR #328, opened 2024-10-03
make it easier to see EMV tags

alovak:
With the change, you don't need to work with the length prefix and separate composite field. Now you can see all EMV tags (that both are in the spec and in the ICC data).


wadearnold (approved):
So Beutiful!

adamdecaf (approved):

PR #334, opened 2024-11-07
Add ability to remove Subfields from a Composite field

Stashchenko:
We have a challenging task of adding and cleaning up `subelements`. Adding a field works fine, but removing one is a bit tricky.

Currently, I’ve found two workarounds:
 - Using Marshal/Unmarshal with a cleaned-up struct.
 - Getting the subfields, removing the unneeded ones, packing back to bytes, and then unpacking into a original composite field.

Both methods are slow and somewhat complex.

This new feature allows for removing subelements in constant time (O(1)) by simply checking the `setSubfields` and deleting the field if it exists.

### DIFF
--- a/field/composite.go
+++ b/field/composite.go
@@ -125,6 +125,20 @@ func (f *Composite) GetSubfields() map[string]Field {
 	return f.getSubfields()
 }
 
+// RemoveSubfield removed subfield by a tag name
+func (f *Composite) RemoveSubfield(indexTag string) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if _, set := f.setSubfields[indexTag]; !set {
+		return errors.New("fieldTag '" + indexTag + "' is not present")
+	}
+
+	delete(f.setSubfields, indexTag)
+
+	return nil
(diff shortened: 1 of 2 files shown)


alovak:
Have you tried to use `err = message.UnsetFields("2", "3.3")`? You can pass path to the subfiled to remove it.

Stashchenko:
> UnsetFields

hm... didn't know about this method. Let me check

alovak:
@Stashchenko I invite you to join Moov Community Slack and there is #iso8583 channel where you can ask question, share feedback. We can chat there too :D

here is the link to join: https://moov-io.slack.com/archives/C014UT7C3ST

Stashchenko:
> @Stashchenko I invite you to join Moov Community Slack and there is #iso8583 channel where you can ask question, share feedback. We can chat there too :D
> 
> here is the link to join: https://moov-io.slack.com/archives/C014UT7C3ST

Thanks @alovak ! I've tested all changes with `UnsetFields` - It works great! Exactly what we need ! I'll close the PR

Issue #327, opened 2024-09-29
odd value parsing issue in ISOMessage


anilgorgec:
Hello,

I have a question regarding odd numbers. for example  the Track 2 Data field has a length of 37 characters, which is an odd number. Since ISO 8583 messages are often processed in hexadecimal or BCD format, they need to be byte-aligned. 

I tried to parse the iso message in the example below.but I could not get the output as expected output.

Can you help me with this issue?

ISO Message:
`020030200580208000000000000000000015001079890071002100375111111211111111d11110000000000000000f5445535430303031`

 output  
```
MTI..........: 0200
Bitmap HEX...: 3020058020800000
Bitmap bits..:
    [1-8]00110000    [9-16]00100000   [17-24]00000101   [25-32]10000000
  [33-40]00100000   [41-48]10000000   [49-56]00000000   [57-64]00000000
F0   Message Type Indicator.................: 0200
F3   Processing Code........................: 0
F4   Transaction Amount.....................: 1500
F11  Systems Trace Audit Number (STAN)......: 107989
F22  Point of Sale (POS) Entry Mode.........: 007
F24  Function Code..........................: 100
F25  Point of Service Condition Code........: 21
F35  Track 2 Data...........................: =^^
F41  Card Acceptor Terminal Identification..: 37511111
``` 

expected output 
```
MTI..........: 0200
Bitmap HEX...: 3020058020800000
Bitmap bits..:
    [1-8]00110000    [9-16]00100000   [17-24]00000101   [25-32]10000000
  [33-40]00100000   [41-48]10000000   [49-56]00000000   [57-64]00000000
F0   Message Type Indicator.................: 0200
F3   Processing Code........................: 0
F4   Transaction Amount.....................: 1500
F11  Systems Trace Audit Number (STAN)......: 107989
F22  Point of Sale (POS) Entry Mode.........: 0071
F24  Function Code..........................: 0021
F25  Point of Service Condition Code........: 00
F35  Track 2 Data...........................: 5111111211111111d11110000000000000000
F41  Card Acceptor Terminal Identification..: TEST0001 ```

alovak:
Could you please share your (Go) spec here?

carlosmgc2003:
I've had exactly this problem and solved it by creating a custom Hex field and a custom BCD prefix:
```go
		35: custom.NewTrack2(&field.Spec{
			Length:      37,
			Description: "Track 2 Data",
			Enc:         encoding.Binary,
[...]

PR #336, opened 2024-11-11
Enhance Track2 Data Implementation

bpross:
This adds:
- `NewTrack1Value` and `NewTrack2Value` functions to construct track1 and track2 data. This mirrors what we have implemented for other fields
- Adds option functions to above constructors to change the default values during packing
- Adds `Track2Packer` and `Track2Unpacker` to handle some custom packing/unpacking for certain networks
- Adds tests for the packing logic

This fixes:
- a few comment errors I noticed while in the code base

### DIFF
--- a/field/packer_unpacker.go
+++ b/field/packer_unpacker.go
@@ -90,16 +90,10 @@ func (p Track2Unpacker) Unpack(packedFieldValue []byte, spec *Spec) ([]byte, int
 		return nil, 0, fmt.Errorf("failed to decode length: %w", err)
 	}
 
-	// Peek and see if the first value is equal to the padding rune if set
-	// if it is, we need to increase the length to decode by 1
-	// This is because the packer will pad the value, but set the value length
-	// equal to the unpadded value length
-	if spec.Pad != nil {
-		b := spec.Pad.Inspect()
-		r := b[0]
-		if packedFieldValue[prefBytes] == r {
-			valueLength++
-		}
+	// if valueLength is odd we need to make it even to adjust for
+	// the padding in our Packer
+	if valueLength%2 != 0 {
+		valueLength++
 	}
 
 	// decode the value
--- a/field/packer_unpacker.go
+++ b/field/packer_unpacker.go
@@ -51,3 +51,67 @@ func (u defaultUnpacker) Unpack(packedFieldValue []byte, spec *Spec) ([]byte, in
 
 	return value, read + prefBytes, nil
 }
+
+type Track2Packer struct{}
+
+// This is a custom packer for Track2 Data. Some specifications require the length
+// to be equal to the length of the pre-padded value.
+func (p Track2Packer) Pack(value []byte, spec *Spec) ([]byte, error) {
+	data := value
+
+	// Only pad if the length is odd. If so, just add
+	// one pad character, so tell the Pad function that
+	// the length we want is +1 to what the value is
+	if spec.Pad != nil && len(value)%2 != 0 {
+		data = spec.Pad.Pad(data, len(value)+1)
+	}
+
+	packed, err := spec.Enc.Encode(data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode content: %w", err)
+	}
+
+	// Encode the length to that of the original string, not the potentially
+	// padded length
+	packedLength, err := spec.Pref.EncodeLength(spec.Length, len(value))
+	if err != nil {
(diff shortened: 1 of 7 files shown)


bpross:
Potentially fixes: https://github.com/moov-io/iso8583/issues/327

field/packer_unpacker.go
```diff
@@ -51,3 +51,67 @@ func (u defaultUnpacker) Unpack(packedFieldValue []byte, spec *Spec) ([]byte, in
...
+	valueLength, prefBytes, err := spec.Pref.DecodeLength(spec.Length, packedFieldValue)
+	if err != nil {
+		return nil, 0, fmt.Errorf("failed to decode length: %w", err)
+	}
+
+	// Peek and see if the first value is equal to the padding rune if set
+	// if it is, we need to increase the length to decode by 1
+	// This is because the packer will pad the value, but set the value length
+	// equal to the unpadded value length
+	if spec.Pad != nil {
```
alovak, line 97: Can we `valueLength++` if the length is odd without checking for the pad char/rune?

field/packer_unpacker.go
```diff
@@ -51,3 +51,67 @@ func (u defaultUnpacker) Unpack(packedFieldValue []byte, spec *Spec) ([]byte, in
...
+	// padded length
+	packedLength, err := spec.Pref.EncodeLength(spec.Length, len(value))
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode length: %w", err)
+	}
+
+	return append(packedLength, packed...), nil
+}
+
+type Track2Unpacker struct{}
```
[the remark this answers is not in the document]
alovak: You know, I'm not sure it will make any difference. The question is: how would someone look for such a packer/unpacker?

field/track2.go
```diff
@@ -29,12 +29,41 @@ const (
...
 		spec: spec,
 	}
 }
 
+func NewTrack2Value(
+	primaryAccountNumber string,
+	expirationDate *time.Time,
+	serviceCode,
+	discretionaryData string,
+	opts ...Track2Option,
```
alovak, line 51: why we can't accept the separator (or format code in case of track 1) as a parameter? It looks like any other field here.

field/track_test.go
```diff
@@ -136,17 +138,17 @@ func TestTrack1(t *testing.T) {
...
 		})
 
 		t.Run("Unmarshal gets track values into data parameter", func(t *testing.T) {
 			expDate, err := time.Parse("0601", "9901")
 			require.NoError(t, err)
 
-			track := NewTrack1(track1Spec)
-			err = track.Marshal(&Track1{
+			track := field.NewTrack1(track1Spec)
+			err = track.Marshal(&field.Track1{
```
alovak, line 151: Should we use NewTrack1Value here instead of struct?

field/packer_unpacker.go
bpross: Yes, now that I think about it we can just check the length instead of the first rune

field/packer_unpacker.go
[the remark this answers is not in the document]
bpross: Would it make more sense to have a `packer` package?

field/track2.go
bpross: For track2 I left `separator` as an option, because there is a default value that is valid for some networks/acquirers. So there exists situations where the caller wouldn't need to set the separator.

For track1, similar reasons. `FormatCode` and `FixedLength` are not needed for all networks/acquirers.

I think this approach makes the call site a bit cleaner:

```
Without options
fd := NewTrack1Value(
  false,
  "",
  "11111...",
  "Ben/Ross",
  "2411",
  "111",
  "1234...",
)
```

```
With options
fd := NewTrack1Value(
  "11111...",
  "Ben/Ross",
  "2411",
  "111",
  "1234...",
)

or

fd := NewTrack1Value(
  "11111...",
  "Ben/Ross",
  "2411",
  "111",
  "1234...",
  WithFixedLength(),
  WithFormatCode("X"),
)
```

bpross: I'm not set on this approach, so if you feel strongly I can just make them params to the constructor.

field/packer_unpacker.go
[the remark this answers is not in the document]
alovak: I would try to avoid it for as long as possible. The packer for track2 is the first candidate for such, but let's wait for some more. For now we can keep it here or move into track2.go.

alovak (approved):

PR #344, opened 2025-01-29
Support marshaling named string types

Inuart:
IMHO types like `type myString string` should also be considered as a `string` when marshaling. Unmarshaling already does this.


alovak (approved):
@Inuart thank you a lot for noticing and fixing it!

alovak:
the change is released -> https://github.com/moov-io/iso8583/releases/tag/v0.23.2

PR #346, opened 2025-01-30
Composite.Unpack shouldn't err if its bitmap is empty

Inuart:
I've encountered messages in the wild that seem to have a composite field present but with length 0. This PR is my best attempt at recreating one in a test and fixing it too.

You can probably write a better test case inside `composite_test.go` but I wasn't able to.


Inuart:
meh, in the end it was another field that was wrong,  sorry for the noise 🙏

PR #355, opened 2025-03-21
Field 54 in the ISO SPEC is 120 length `an LLL`

parsnips:
The fields beneath it are fixed 20 length additional amounts.

I've tried to write a composite field for it, but if the field is present in the bitmap and zero length the library fails with:

```
failed to unpack message: failed to unpack field 54 (Additional
Amounts): failed to unpack subfield 01: not enough data to unpack,
expected: 20, got: 0
```

This patch "fixes" that, by just returning the offset if there's no data in a composite spec.. But i'm not certain this is the correct fix. (tests all pass, including these).


parsnips:
@Inuart pointed out that he ran into a similar situation with an earlier field in the spec being incorrect.  Ran my results via `iso8583.DescribeMessage` and saw that the `Transaction Currency Code` was showing up as `000` instead of `840`... and figured out it was field 43 with the problem:

```
MTI..........: 0120
Bitmap HEX...: F67A44D108E0A40A0000000004020000
Bitmap bits..:
    [1-8]11110110    [9-16]01111010   [17-24]01000100   [25-32]11010001
  [33-40]00001000   [41-48]11100000   [49-56]10100100   [57-64]00001010
  [65-72]00000000   [73-80]00000000   [81-88]00000000   [89-96]00000000
 [97-104]00000100 [105-112]00000010 [113-120]00000000 [121-128]00000000
F0   Message Type Indicator.....................: 0120
F2   Primary Account Number.....................: 1001****6564
F3   Processing Code............................: 0
F4   Transaction Amount.........................: 50000
F6   Billing Amount.............................: 50100
F7   Transmission Date & Time...................: 0122133300
F10  Cardholder Billing Conversion Rate.........: 61002000
F11  Systems Trace Audit Number (STAN)..........: 016374
F12  Local Transaction Time.....................: 183300
F13  Local Transaction Date.....................: 0122
F15  Settlement Date............................: 0122
F18  Merchant Type..............................: 5541
F22  Point of Sale (POS) Entry Mode.............: 902
F25  Point of Service Condition Code............: 00
F26  Point of Service PIN Capture Code..........: 00
F28  Transaction Fee Amount.....................: D00000000
F32  Acquiring Institution Identification Code..: 000000
F37  Retrieval Reference Number.................: 802213016374
F41  Card Acceptor Terminal Identification......: TERMID01
F42  Card Acceptor Identification Code..........: CARD ACCEPTOR  
F43  Card Acceptor Name/Location................: ACQUIRER NAME            CITY NAME    CAUSA8408400200072840D0000000001
F49  Transaction Currency Code..................: 000
F51  Cardholder Billing Currency Code...........: 160
F54  Additional Amounts SUBFIELDS:
-------------------------------------------
------------------------------------------
F61  Point of Service Data.....: 
F63  Network Data..............: 0200
F102 ACCOUNT IDENTIFICATION 1..: 
F111 ADDITIONAL DATA...........: 0002    123456123456123           0                       VISA       
```


The spec involved for the company involved (i2c) has 43 as possibly 3 different values based on config:
- `an 43` - described as for "existing clients"
- `an 70` - described as for "new clients"
- `an lll 255` - so new, they haven't described who it is for

I made the spec _assuming_ the examples would be oriented to new clients... but looks like that was a bad assumption :)

PR #371, opened 2025-06-27
fix: a binary field can marshal to a named string

Inuart:
Similar to https://github.com/moov-io/iso8583/pull/344 but this time for when the spec field is a Binary and has to Marshal into a named string.

### DIFF
--- a/message_test.go
+++ b/message_test.go
@@ -2279,6 +2279,17 @@ func TestStructWithTypes(t *testing.T) {
 				},
 				expectedPackedString: "011040000000000000000000000000000000164242424242424242",
 			},
+			{
+				name: "struct with named string type and value set on a binary field",
+				input: struct {
+					MTI                  myString `index:"0"`
+					PrimaryAccountNumber myString `index:"4"`
+				}{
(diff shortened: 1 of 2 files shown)


message_test.go
```diff
@@ -2279,6 +2279,17 @@ func TestStructWithTypes(t *testing.T) {
...
 				expectedPackedString: "011040000000000000000000000000000000164242424242424242",
 			},
+			{
+				name: "struct with named string type and value set on a binary field",
+				input: struct {
+					MTI                  myString `index:"0"`
+					PrimaryAccountNumber myString `index:"4"`
+				}{
+					MTI:                  "0110",
+					PrimaryAccountNumber: myString("313233"), // "123" in Hex
```
alovak, line 2289: should it be:
```suggestion
					BinaryData myString `index:"4"`
				}{
					MTI:                  "0110",
					BinaryData: myString("313233"), // "123" in Hex
```

Inuart: done!

alovak (approved):

PR #375, opened 2025-08-01
change some error messages and wrap prefix error into prefix.LengthError type

alovak:
Changes:
* change error message to be more specific about data, not field
* add prefix.LengthError type

This change shouldn't break anything, unless you rely on message text.


adamdecaf (approved):

adamdecaf:
This will most likely break folks, so it should be released in a minor release - not bugfix.

PR #379, opened 2025-08-09
feat(encoding): add nibble‑aware Track2BCD encoder/decoder for DE35

trylix:
### Summary

- Introduces encoding.Track2BCD, a nibble‑aware encoder/decoder for packed BCD Track‑2 data.
- Interprets length as number of digits (nibbles) and consumes exactly ceil(n/2) bytes.
- Handles odd lengths with the leading‑zero rule.
- Validates the Track‑2 nibble set (0–9 and separator nibble 0xD).
- Compatible with prefix.Binary.L using a light Track‑2 packer/unpacker (optional).

> This is additive and opt‑in; existing encoders/prefixers continue to work as before.

### Motivation & context

Per Visa’s F35 spec, the length byte is the count of hex digits (nibbles), not bytes; content is 4‑bit BCD (unsigned packed), and odd digit counts use a leading zero in the first unused half‑byte. Track‑2 also includes a separator nibble 0xD.
Generic ASCII/hex decoders don’t match these semantics (they either over‑read bytes or fail on non‑ASCII 0xD), so parsing desynchronizes.

### What’s included

- `encoding.Track2BCD` (encoder/decoder).
- Unit tests: even/odd lengths, separator, invalid nibbles, round‑trip.

### Backward compatibility

- No breaking changes.
- New type is opt‑in. Existing field specs and encoders remain untouched.

### DIFF
--- a/encoding/track2_bcd.go
+++ b/encoding/track2_bcd.go
@@ -1,0 +1,84 @@
+package encoding
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/yerden/go-util/bcd"
+
+	"github.com/moov-io/iso8583/utils"
+)
+
+// trackDigitsBCD defines the BCD mapping for digits '0'-'9' and the separator 'D'.
+var trackDigitsBCD = &bcd.BCD{
+	Map: func() map[byte]byte {
+		m := make(map[byte]byte, 11)
+		for c := byte('0'); c <= byte('9'); c++ {
+			m[c] = c - '0'
+		}
+		m['D'] = 0xD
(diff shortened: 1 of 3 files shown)


alovak:
Hey! Thank you for the PR and effort to address the issue.

I'm concerned that in this PR we add encoding and not a custom field type e.g. field.Track2 or some packer/unpacker for Track2. The logic that we have for Track2 / Visa is field specific, not encoding specific.

How would you address the same issue with introducing custom field/packer/unpacker or modifying existing `field.Track2`?

PR #400, opened 2026-01-15
feat: handle multiple occurrences of the same data fields

EngMahmut:
### **Bug Fix Summary**
**Problem**
When a composite field contained the same tag (e.g., "1a") appearing multiple times, only the last occurrence was kept - previous values were overwritten and data was lost.

**Example:**
Input:  Tag "1a" with value "ABC"
        Tag "1a" with value "DEF"
        Tag "1a" with value "GHI"

**Before Fix:**
  subfields["1a"] = "GHI"  (only last value, data loss!)

**After Fix:**
  subfields["1a"]   = "ABC" (first occurrence)
  subfields["1a_1"] = "DEF" (second occurrence with suffix)
  subfields["1a_2"] = "GHI" (third occurrence with suffix)
  
###   Solution
Added two helper functions and updated the unpacking/packing logic in composite.go:

[getUniqueFieldKey(tag)](vscode-file://vscode-app/c:/Users/HP/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - When unpacking duplicate tags, generates unique keys with numeric suffixes:

First occurrence: "1a"
Second occurrence: "1a_1"
Third occurrence: "1a_2", etc.
[extractBaseTag(fieldKey)](vscode-file://vscode-app/c:/Users/HP/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) - When packing, removes the suffix to restore the original tag:

"1a" → "1a"
"1a_1" → "1a"
"1a_2" → "1a", etc.
Updated unpackSubfieldsByTag() - Uses getUniqueFieldKey() to preserve all duplicate occurrences

Updated packByTag() - Uses [extractBaseTag()](vscode-file://vscode-app/c:/Users/HP/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to restore original tag names when packing

Test Results
✅ All tests PASS:

single_occurrence - Basic functionality
multiple_tags_no_data_loss - Tag-based composites
pack_unpack_consistency - Round-trip encoding/decoding
same_tag_multiple_occurrences_bug_repro - BUG FIXED! 🎉
All 3 occurrences of tag "1a" are now correctly preserved with their distinct values!

### DIFF
--- a/field/composite.go
+++ b/field/composite.go
@@ -496,8 +496,12 @@ func (f *Composite) packByTag() ([]byte, error) {
 		return nil, errors.New("cannot pack composite field by tag when Tag spec is not defined")
 	}
 
-	for _, tag := range orderedKeys(f.subfields, f.spec.Tag.Sort) {
-		field := f.subfields[tag]
+	for _, fieldKey := range orderedKeys(f.subfields, f.spec.Tag.Sort) {
+		field := f.subfields[fieldKey]
+
+		// Extract the base tag name (remove suffix if exists)
+		// e.g., "1a_1" -> "1a", "1a_2" -> "1a", "1a" -> "1a"
+		tag := extractBaseTag(fieldKey)
 
 		if f.spec.Tag.Enc != nil {
 			tagBytes := []byte(tag)
@@ -515,7 +519,7 @@ func (f *Composite) packByTag() ([]byte, error) {
 
 		packedBytes, err := field.Pack()
 		if err != nil {
-			return nil, fmt.Errorf("failed to pack subfield %v: %w", tag, err)
+			return nil, fmt.Errorf("failed to pack subfield %v: %w", fieldKey, err)
 		}
 
 		packed = append(packed, packedBytes...)
@@ -690,7 +694,9 @@ func (f *Composite) unpackSubfieldsByTag(data []byte) (int, string, error) {
 					if err := binaryField.SetBytes(fieldData); err != nil {
 						return 0, tag, fmt.Errorf("failed to set bytes for unknown tag %s: %w", tag, err)
 					}
-					f.subfields[tag] = binaryField
+					// Handle duplicate unknown tags with suffix
+					fieldKey := f.getUniqueFieldKey(tag)
+					f.subfields[fieldKey] = binaryField
 				}
 
 				offset += fieldLength + read
@@ -702,7 +708,9 @@ func (f *Composite) unpackSubfieldsByTag(data []byte) (int, string, error) {
 		}
 
 		field := NewInstanceOf(specField)
-		f.subfields[tag] = field
+		// Handle duplicate tags by adding numeric suffixes
+		fieldKey := f.getUniqueFieldKey(tag)
+		f.subfields[fieldKey] = field
 
 		read, err = field.Unpack(data[offset:])
 		if err != nil {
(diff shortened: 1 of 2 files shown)


andresantoniuk:
I pulled this branch and ran it against a BER-TLV composite, since it touches something I depend on. It fixes a real loss, and I found a couple of things while testing that may be worth knowing.

**Before and after.** Same spec, same input: six data objects, `9F02` three times and `9A` twice, interleaved the way a terminal would emit them.

```go
// field/probe_test.go, in package field
spec := &Spec{
	Length: 999, Description: "ICC", Pref: prefix.ASCII.LLL,
	Tag: &TagSpec{Enc: encoding.BerTLVTag, Sort: sort.StringsByHex},
	Subfields: map[string]Field{
		"9A":   NewHex(&Spec{Description: "Transaction Date", Enc: encoding.Binary, Pref: prefix.BerTLV}),
		"9F02": NewHex(&Spec{Description: "Amount", Enc: encoding.Binary, Pref: prefix.BerTLV}),
		"5F2A": NewHex(&Spec{Description: "Currency", Enc: encoding.Binary, Pref: prefix.BerTLV}),
	},
}

content, _ := hex.DecodeString(
	"9F0206000000000501" + "9A03260824" + "5F2A020858" +
		"9F0206000000000999" + "9A03260825" + "9F0206000000000777")
in := append([]byte(fmt.Sprintf("%03d", len(content))), content...)

f := NewComposite(spec)
f.Unpack(in)
out, _ := f.Pack()
```

```
in       = 3034329f02060000000005019a032608245f2a0208589f02060000000009999a032608259f0206000000000777
master   = 3031399a032608255f2a0208589f0206000000000777
this PR  = 3034329a032608245f2a0208589a032608259f02060000000005019f02060000000009999f0206000000000777
```

Read as tags:

```
in       9F02(0501)  9A(260824)  5F2A(0858)  9F02(0999)  9A(260825)  9F02(0777)
master   9A(260825)  5F2A(0858)  9F02(0777)
this PR  9A(260824)  5F2A(0858)  9A(260825)  9F02(0501)  9F02(0999)  9F02(0777)
```

On master three of the six objects are silently gone, 45 bytes in and 22 out. On this branch all six survive with their values, and the repeated ones keep their relative order, which is the half that matters. It is deterministic as well: 300 unpack/pack cycles produced a single output.

**On your open question about packing repeated data:** the order that comes back is the sorter's rather than the message's. Everything is regrouped by tag, and the two `9A` occurrences come back separated by `5F2A`.

**Why they end up separated:** the suffixed key is sorted on its own merits. `StringsByHex` cannot hex-decode `9A_1`, so for any pair involving it the comparator falls back to `x[i] < x[j]`, while other pairs are compared as big-endian integers. Mixing two criteria inside one `Less` is not something `sort.Slice` promises anything about. It was stable across my runs, but `9A_1` lands wherever its string sorts rather than beside `9A`.

**The shape underneath both:** a map cannot express duplicates, and it cannot express order either, so `Unpack` followed by `Pack` does not reproduce the bytes it was given even without duplicates. `moov-io/bertlv` already models this same data as an ordered [`[]TLV`](https://github.com/moov-io/bertlv/blob/ce0986add357088ad5af8cf971c721e308c4fec9/tlv.go#L13-L16), and [`BuildTagMap`](https://github.com/moov-io/bertlv/blob/ce0986add357088ad5af8cf971c721e308c4fec9/optimization.go#L26-L29) builds `map[string][]TLV` on demand when lookup is needed, noting that duplicates are "preserved in the order they are encountered". An ordered view carried alongside `subfields`, with `Sort` still deciding anything that was not unpacked, would cover both halves here without changing how subfields are looked up.

For context on why the ordering half matters to me: my own use is a protocol-aware load balancer. On legs that only read a message `MessageScanner` covers it, but legs that add or compute a field have to re-encode, and then a TLV field nobody touched comes back reordered as a side effect of a change made somewhere else.

Happy to share the tests, and to help with either half.

Issue #410, opened 2026-04-01
`Clone()` in `message.go` introduces a Double-Lock Deadlock Risk


ashwingopalsamy:
`Message.Clone()` is implemented via a pack -> unpack round-trip. 

This is semantically incorrect: fields explicitly set to a zero value (e.g. Processing Code "`000000`") are silently dropped from the clone because `fieldsMap` is rebuilt from bitmap during `Unpack`, and zero-value fields may not have their bitmap bits set depending on spec configuration.

### Root Cause

```go
func (m *Message) Clone() (*Message, error) {
    // ...
    bytes, err := m.wrapErrorPack()    // bitmap is set based on fieldsMap
    // ...
    newMessage.Unpack(bytes)           // fieldsMap rebuilt from bitmap (lossy)
    // ...
}
```

This is a real risk in reversal and advice flows, where a **0100** is cloned and mutated into a **0400**; directly resulting in producing a structurally incorrect financial transaction message.

ashwingopalsamy:
Added a pull #411 that directly addresses this issue. Kindly review.

alovak:
Could you please provide an example? I'm not sure I understand the issue. If we pack the field and then unpack it, how do we create and then drop zero value fields?

PR #411, opened 2026-04-01
fix: replace Clone() binary round-trip with direct field copy

ashwingopalsamy:
### Changes

This PR directly addresses the gap identified with Issue #410 . By replacing the round-trip with a direct copy; each field's decoded value bytes are copied individually, and `fieldsMap` membership is preserved exactly as it was on the original.

All existing tests pass; added a new test for the proposed fix, including a `pack -> unpack-> unmarshal` round-trip to confirm the fields survive the wire format accordingly.

### DIFF
--- a/message.go
+++ b/message.go
@@ -377,33 +377,35 @@ func (m *Message) packableFieldIDs() ([]int, error) {
 	return populatedFieldIDs, nil
 }
 
-// Clone clones the message by creating a new message from the binary
-// representation of the original message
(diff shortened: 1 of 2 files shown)


alovak:
@ashwingopalsamy thank you for the PR. It looks good. I'll check some aspects of it. Do you have any benchmarks to compare how faster / better it is now? If no, could you please add one for OldClone and NewClone (to the PR description) and you can use our setup we have here https://github.com/moov-io/iso8583/blob/master/benchmarks_test.go.

PR #412, opened 2026-04-10
feat: add support for re-marshaling unknown TLV tags in MarshalPath

MassimilianoPoggiQonto:
Currently, we have the option to parse and unknown TLV tags from a message.
However, once the message is un-marshaled into a struct, there is no way to re-add those fields back when re-marshaling from the struct.

The only current approach I could find would require keeping a reference to the entire message, and re-marshaling the (eventually modified) struct into the original message.

What I'm proposing here is an extension of the `MarshalPath` method for composite fields: if a composite field allows parsing and storing unknown TLV tags, then its `MarshalPath` implementation will allow directly setting a value provided it satisfies the `Field` interface if no definition is found in the spec.

Additionally, I've added an utility method to retrieve unknown TLVs: for our usage of the library, we sometimes delay parsing some data elements until necessary.
This means that we sometimes don't parse the entire message, but just a `Composite` spec for the specific data elements.

### DIFF
--- a/field/composite.go
+++ b/field/composite.go
@@ -806,6 +806,15 @@ func (m *Composite) MarshalPath(path string, value any) error {
 
 	field, err := m.getOrCreateField(id)
 	if err != nil {
+		// at the moment getOrCreateField only returns an error if the field is not in spec
+		// if the passed value is of type Field, and there is no subPath, we can set it directly without returning an error
+		// this allows us to restore subfields corresponding to unknown TLV tags
+		if v, ok := value.(Field); !hasSubPath && m.skipUnknownTLVTags() &&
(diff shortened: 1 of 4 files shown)


alovak:
@MassimilianoPoggiQonto thank you for the PR! It looks good! Could you please add tests for the `UnknownCompositeTags` helper, once added we can merge it.

P.S. I pushed the typo commit to your branch.

MassimilianoPoggiQonto:
Hey @alovak, thanks for fixing the typo!
I've added the missing tests (hope you don't mind, I borrowed the spec and test data from the one above 😅).

alovak (approved):

PR #421, opened 2026-06-25
feat(composite): add TagSpec.Prefix to support Tag Class Character (TCC)

frochyzhang:
## Summary

Some ISO 8583 implementations, such as **Mastercard DE 48**, use a Tag Class Character (TCC) byte (e.g., EBCDIC `T` = `0xE3`) that precedes all tag-value pairs in a composite field. The TCC indicates the tag encoding scheme for the entire TLV structure.

Previously, there was no way to handle this at the framework level — users had to manually prepend/strip the TCC byte outside of the Composite field, breaking encapsulation.

## Changes

- Add `Prefix []byte` field to `TagSpec` for wire-format bytes before tag-value pairs
- `packByTag()`: write prefix before the first tag-value pair
- `unpackSubfieldsByTag()`: read and validate prefix before decoding tags
- Add test coverage: pack, unpack, wrong prefix error, too-short data, backward compatibility (prefix=nil)

## Usage

```go
Spec{
    Tag: &TagSpec{
        Length: 2,               // fixed EBCDIC tag length
        Enc:    encoding.EBCDIC, // EBCDIC tag encoding
        Sort:   sort.StringsByInt,
        Prefix: []byte{0xE3},   // EBCDIC "T" — TCC
    },
    Subfields: map[string]Field{...},
}
```

## Checklist

- [x] `go build` passes
- [x] `go test ./...` passes (all 13 packages)
- [x] `go vet ./field/...` clean
- [x] `gofmt` no diffs
- [x] `gitleaks` no leaks
- [x] Test coverage: 84.0% (threshold: 75%)
- [x] Backward compatible (prefix=nil preserves existing behavior)

### DIFF
--- a/field/composite.go
+++ b/field/composite.go
@@ -1,6 +1,7 @@
 package field
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -500,6 +501,11 @@ func (f *Composite) packByTag() ([]byte, error) {
 		return nil, errors.New("cannot pack composite field by tag when Tag spec is not defined")
 	}
 
+	// prepend tag prefix (e.g. TCC byte) only if there are subfields to pack
+	if len(f.spec.Tag.Prefix) > 0 && len(f.subfields) > 0 {
+		packed = append(packed, f.spec.Tag.Prefix...)
+	}
+
 	for _, tag := range orderedKeys(f.subfields, f.spec.Tag.Sort) {
 		field := f.subfields[tag]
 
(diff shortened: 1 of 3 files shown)


alovak:
Sorry for the late reply. With the proposed approach the TCC is just dropped during parsing, but I see that TLVs may be encoded based on the TCC value.

There is an alternative way that may work for you. Each TCC can be its own tag in the specification, then such tag will have subfields (TLVs). Something like that:

```
fieldSpec = &field.Spec{
	Length:      1535,
	Description: "Data",
	Pref:        prefix.Binary.LL,
	Tag: &field.TagSpec{
		Length: 1, // that's your TCC
		Enc:    encoding.ASCIIHexToBytes, // to work with 0xE3
		Sort:   sort.StringsByHex,
		// we want to skip unknown TCCs
		SkipUnknownTLVTags: true,
		PrefUnknownTLV:     prefix.Binary.LL, // length of the prefix after TCC (maybe None will work for your case)
	},
	Subfields: map[string]field.Field{
		"E3": field.NewComposite(&field.Spec{ // DataSet ID
			Length:      1000, // for None prefix all data will be passed to the composite to unpack
			Description: "TCC T Data",
			Pref:        prefix.Binary.None,
			Tag: &field.TagSpec{
				Enc:                encoding.BerTLVTag,
				Sort:               sort.StringsByHex,
				SkipUnknownTLVTags: true,
			},
			Subfields: map[string]field.Field{
				"86": field.NewString(&field.Spec{
					Length:      8,
					Description: "Test 2",
					Enc:         encoding.EBCDIC,
					Pref:        prefix.BerTLV,
				}),
				"89": field.NewString(&field.Spec{
					Length:      45,
					Description: "Test 1",
					Enc:         encoding.EBCDIC,
					Pref:        prefix.BerTLV,
				}),



```

I see that there is not lenght prefix after TCC, but maybe prefix.None will work here. Let me know if something like that works for you.

PR #453, opened 2026-09-14
ci: track marshal and unpack Go benchmarks

adamdecaf:
## Why

`BenchmarkMarshaling` and `BenchmarkUnpacking` are the useful hot path in this module. They were not on the Moov hub. Running them here means the chart commit is an iso8583 SHA.

## What

- `make bench` runs those two tests with `-benchmem`.
- Nightly (and `workflow_dispatch`) stores `docs/bench`.
- `push` CI ignores `docs/bench/**`.

The hub at https://moov-io.github.io/benchmarks/ will pick up the file after the first run.

🤖 grok


adamdecaf:
Hub change that consumes `docs/bench/data.js` after this lands: https://github.com/moov-io/benchmarks/pull/18

🤖 grok

